PR #2, opened 2024-08-08
IOTAERP 8

jcbbb:
I want to insert user role to user_roles table when creating a user. What is the order of operations for this? Do I call RoleService from UserService.Create or do everything in controller? What was your idea here for this? Or should I be doing this in repository?

### DIFF
--- a/internal/domain/user/user.go
+++ b/internal/domain/user/user.go
@@ -4,11 +4,18 @@ import (
 	"strings"
 	"time"
 
+	"github.com/iota-agency/iota-erp/internal/domain/role"
 	model "github.com/iota-agency/iota-erp/internal/interfaces/graph/gqlmodels"
 	"github.com/iota-agency/iota-erp/sdk/utils/sequence"
 	"github.com/nicksnyder/go-i18n/v2/i18n"
 	"golang.org/x/crypto/bcrypt"
 )
+
+type UserRole struct {
+	RoleId int64
+	UserId int64
+	Role   role.Role
+}
 
 type User struct {
 	Id         int64
@@ -24,6 +31,7 @@ type User struct {
 	LastAction *time.Time
 	CreatedAt  time.Time
 	UpdatedAt  time.Time
+	Roles      []*role.Role `gorm:"many2many:user_roles"`
 }
 
 type UserUpdate struct {
@@ -31,6 +39,7 @@ type UserUpdate struct {
 	LastName  string
 	Email     string
 	Password  string
+	RoleId    int64
 }
 
 func (u *User) CheckPassword(password string) bool {
@@ -93,6 +102,9 @@ func (u *UserUpdate) Ok(l *i18n.Localizer) (map[string]string, bool) {
 	}
 	if strings.TrimSpace(u.FirstName) == "" {
 		errors["email"] = l.MustLocalize(&i18n.LocalizeConfig{MessageID: "Validations.Required"})
+	}
+	if u.RoleId == 0 {
+		errors["roleId"] = l.MustLocalize(&i18n.LocalizeConfig{MessageID: "Validations.Required"})
 	}
 	return errors, len(errors) == 0
 }
--- a/internal/presentation/controllers/user_controller.go
+++ b/internal/presentation/controllers/user_controller.go
@@ -7,6 +7,7 @@ import (
 	"github.com/a-h/templ"
 	"github.com/gorilla/mux"
 	"github.com/iota-agency/iota-erp/internal/app/services"
+	"github.com/iota-agency/iota-erp/internal/domain/role"
 	"github.com/iota-agency/iota-erp/internal/domain/user"
 	"github.com/iota-agency/iota-erp/internal/presentation/templates/pages/users"
 	"github.com/iota-agency/iota-erp/internal/presentation/types"
@@ -92,10 +93,17 @@ func (c *UsersController) PostEdit(w http.ResponseWriter, r *http.Request) {
 	}
 	action := r.FormValue("_action")
 	if action == "save" {
+		var roleId int
+		roleId, err = strconv.Atoi(r.FormValue("roleId"))
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
 		upd := &user.UserUpdate{
 			FirstName: r.FormValue("firstName"),
 			LastName:  r.FormValue("lastName"),
 			Email:     r.FormValue("email"),
+			RoleId:    int64(roleId),
 		}
 		var pageCtx *types.PageContext
 		pageCtx, err = composables.UsePageCtx(r, &composables.PageData{Title: "Users.Meta.Edit.Title"})
@@ -119,7 +127,7 @@ func (c *UsersController) PostEdit(w http.ResponseWriter, r *http.Request) {
 			templ.Handler(users.EditForm(pageCtx.Localizer, us, roles, errors), templ.WithStreaming()).ServeHTTP(w, r)
 			return
 		}
-		err = c.app.UserService.Update(r.Context(), &user.User{Id: int64(id), FirstName: upd.FirstName, LastName: upd.LastName, Email: upd.Email})
+		err = c.app.UserService.Update(r.Context(), &user.User{Id: int64(id), FirstName: upd.FirstName, LastName: upd.LastName, Email: upd.Email, Roles: []*role.Role{{Id: upd.RoleId}}})
 	} else if action == "delete" {
 		_, err = c.app.UserService.Delete(r.Context(), int64(id))
 	}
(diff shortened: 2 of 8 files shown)


docker-compose.dev.yml
```diff
@@ -2,7 +2,7 @@ version: '3.8'
 
 services:
   db:
-    image: postgres:15.1
+    image: ankane/pgvector:v0.5.1
```
diyor28: Roll this change back. All of the pgvector logic will be part of py-embed service & a separate DB. Remove any pgvector extension lines from migrations if needed

internal/configuration/configuration.go
```diff
@@ -66,5 +69,6 @@ func (c *Configuration) Load() error {
 	c.OpenAIKey = env.MustGetEnv("OPENAI_KEY")
 	c.UploadsPath = env.GetEnv("UPLOADS_PATH", "uploads")
 	c.FrontendDomain = env.GetEnv("FRONTEND_DOMAIN", "localhost")
+	c.PageSize, _ = strconv.Atoi(env.GetEnv("PAGE_SIZE", "25"))
```
diyor28: Proper error handling

internal/domain/user/user.go
```diff
@@ -55,3 +73,10 @@ func (u *User) ToGraph() *model.User {
 		UpdatedAt:  u.UpdatedAt,
 	}
 }
+
+func pad(b *bytes.Buffer, str string) {
```
diyor28: make this a part of sdk/utils/sequences

internal/presentation/controllers/user_controller.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/gorilla/mux"
+	"github.com/iota-agency/iota-erp/internal/app/services"
+	"github.com/iota-agency/iota-erp/internal/domain/user"
+	"github.com/iota-agency/iota-erp/internal/presentation/templates/pages/users"
+	"github.com/iota-agency/iota-erp/internal/presentation/types"
+	"github.com/iota-agency/iota-erp/pkg/composables"
+	"github.com/nicksnyder/go-i18n/v2/i18n"
+)
+
+type UserController struct {
```
diyor28: UserController -> UsersController

internal/presentation/controllers/user_controller.go
```diff
@@ -0,0 +1,164 @@
...
+	localizer, found := composables.UseLocalizer(r.Context())
+	if !found {
+		http.Error(w, "localizer not found", http.StatusInternalServerError)
+		return
+	}
+
+	pageCtx := &types.PageContext{
+		Localizer: localizer,
+		Pathname:  pathname,
+		Title:     localizer.MustLocalize(&i18n.LocalizeConfig{MessageID: "NavigationLinks.Users"}),
```
diyor28: For every page add a meta tags sections. Something like
"users": {"meta": {"title": "", "description": ""}}

This way the line turns into:

localizer.MustLocalize(&i18n.LocalizeConfig{MessageID: "Users.Meta.Title"}),

internal/domain/user/user.go
```diff
@@ -39,6 +41,22 @@ func (u *User) SetPassword(password string) error {
 	return nil
 }
 
+func (u *User) FullName() string {
+	out := new(bytes.Buffer)
```
diyor28: just curious, what is the advantage of doing string concatenation using buffers instead of just strings?

internal/presentation/controllers/home_controller.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (c *HomeController) Home(w http.ResponseWriter, r *http.Request) {
+	pathname := r.URL.Path
+	localizer, found := composables.UseLocalizer(r.Context())
+	if !found {
+		http.Error(w, "localizer not found", http.StatusInternalServerError)
+		return
+	}
+	pageCtx := &types.PageContext{
```
diyor28: Would be nice to have a function that constructs page context from *http.Request

internal/presentation/controllers/user_controller.go
```diff
@@ -0,0 +1,164 @@
...
+	pageCtx := &types.PageContext{
+		Localizer: localizer,
+		Pathname:  pathname,
+		Title:     localizer.MustLocalize(&i18n.LocalizeConfig{MessageID: "EditUser"}),
+	}
+	templ.Handler(users.Edit(pageCtx, us, roles), templ.WithStreaming()).ServeHTTP(w, r)
+}
+
+func (c *UserController) DeleteUser(w http.ResponseWriter, r *http.Request) {
+	id, _ := strconv.Atoi(mux.Vars(r)["id"])
```
diyor28: Proper error handling

internal/presentation/controllers/user_controller.go
```diff
@@ -0,0 +1,164 @@
...
+		Pathname:  pathname,
+		Title:     localizer.MustLocalize(&i18n.LocalizeConfig{MessageID: "EditUser"}),
+	}
+
+	w.Header().Add("HX-Push-Url", "/users")
+	templ.Handler(users.UsersContent(pageCtx, us), templ.WithStreaming()).ServeHTTP(w, r)
+}
+
+func (c *UserController) PostEdit(w http.ResponseWriter, r *http.Request) {
+	id, _ := strconv.Atoi(mux.Vars(r)["id"])
```
diyor28: error handling

internal/presentation/controllers/user_controller.go
```diff
@@ -0,0 +1,164 @@
...
+	action := r.FormValue("_action")
+	var err error
+	if action == "save" {
+		err = c.app.UserService.Update(r.Context(), &user.User{
+			Id:        int64(id),
+			FirstName: r.FormValue("firstName"),
```
diyor28, line 103: Adding some validation would be great

internal/presentation/controllers/user_controller.go
```diff
@@ -0,0 +1,164 @@
...
+		Pathname:  pathname,
+		Title:     localizer.MustLocalize(&i18n.LocalizeConfig{MessageID: "NewUser"}),
+	}
+	templ.Handler(users.New(pageCtx, roles), templ.WithStreaming()).ServeHTTP(w, r)
+}
+
+func (c *UserController) CreateUser(w http.ResponseWriter, r *http.Request) {
+	user := &user.User{
+		FirstName: r.FormValue("firstName"),
+		LastName:  r.FormValue("lastName"),
```
diyor28, line 165: Basic validation is a must

internal/presentation/templates/components/base/input/input.templ
```diff
@@ -0,0 +1,104 @@
...
+	class := ""
+	if props.Addon != nil {
+		if props.Addon.Position == AddonLeft {
+			class += "pl-10"
+		} else if props.Addon.Position == AddonRight {
+			class += "pr-10"
+		}
+	}
+
+	return &Props{id: id, Label: props.Label, Class: class + " " + props.Class, Attrs: props.Attrs, Addon: props.Addon, typ: props.typ, Placeholder: props.Placeholder}
```
diyor28: Write an alternative to : https://www.npmjs.com/package/tailwind-merge in Go and use it for class merging

pkg/locales/en.json
```diff
@@ -1,5 +1,48 @@
...
+    "Enums": "Enums",
+    "TaskTypes": "Task Types",
+    "Positions": "Positions",
+    "CashFlow": "Cash Flow",
+    "Projects": "Projects",
+    "ExpenseCategories": "Expense Categories",
+    "Payments": "Payments",
+    "Reports": "Reports"
+  },
+  "EditUser": "Edit user",
```
diyor28: Group messages by page:

Users/Edit/FirstName
Users/List/NewUser

pkg/composables/request.go
```diff
@@ -0,0 +1,28 @@
+package composables
+
+import (
+	"net/http"
+	"strconv"
+)
+
+const pageSize = 25
```
diyor28: Pls make these part of config

internal/domain/user/user.go
jcbbb: As far as I know, it copies previous string to create a new string if you just are using + operator. It probably does not matter for a small string like this, but I got used to doing it like this. Recently I found out about this https://pkg.go.dev/strings#Builder, which could be used instead of `bytes.Buffer`

diyor28: Got it, so it avoids the whole len >= capacity, find a new block of heap memory & copy the data. Though it's not clear to me how it knows the size in advance

internal/presentation/templates/components/base/input/input.templ
jcbbb: Why do we need this? To override classes?

diyor28: Yes and to merge classes properly. Not a high priority task, can be moved to backlog if takes longer than an hour

jcbbb: > Yes and to merge classes properly. Not a high priority task, can be moved to backlog if takes longer than an hour

I can still override with just this, just pass any tailwind class you want. I put base styles in tailwind layers, so it shouldn't conflict

diyor28: What about pr-4 pl-4 turning into px-4 and other similar cases?

jcbbb: For this I suggest just to pass px-4 :)

PR #19, opened 2024-11-25
change upload int references to varchar

jcbbb:


modules/core/migrations/20240602202521-initial.sql
```diff
@@ -3,12 +3,12 @@ BEGIN;
...
 CREATE TABLE uploads
 (
-    id         SERIAL PRIMARY KEY,
-    name       VARCHAR(255) NOT NULL,
-    url        VARCHAR(255) NOT NULL,
---     uploader_id INT          REFERENCES users (id) ON DELETE SET NULL,
-    mimetype   VARCHAR(255) NOT NULL,
-    size       INT          NOT NULL,
+    id         VARCHAR(255) PRIMARY KEY,
+    url        VARCHAR(1024) NOT NULL   DEFAULT '',
```
diyor28, line 9: Are u sure this even works? I think VARCHAR support 255 max, and if you wanna go beyond that you use TEXT

modules/core/migrations/20240602202521-initial.sql
```diff
@@ -3,12 +3,12 @@ BEGIN;
...
+    name       VARCHAR(255)  NOT NULL   DEFAULT '',
```
diyor28, line 10: Also it's better to embed defaults on the Application layer. Makes it more flexible. And DDD suggest to do that as well...

modules/core/migrations/20240602202521-initial.sql
```diff
@@ -3,12 +3,12 @@ BEGIN;
 
 CREATE TABLE uploads
 (
-    id         SERIAL PRIMARY KEY,
-    name       VARCHAR(255) NOT NULL,
-    url        VARCHAR(255) NOT NULL,
---     uploader_id INT          REFERENCES users (id) ON DELETE SET NULL,
-    mimetype   VARCHAR(255) NOT NULL,
-    size       INT          NOT NULL,
+    id         VARCHAR(255) PRIMARY KEY,
```
diyor28: Why not uuid?

diyor28 (changes requested):

modules/core/migrations/20240602202521-initial.sql
jcbbb: > Why not uuid?

ID is actually a file hash to avoid duplicate files, now I think it was a bad idea because of indexing

modules/core/migrations/20240602202521-initial.sql
jcbbb: Yeah, I think it goes up to ~65K characters

modules/core/migrations/20240602202521-initial.sql
jcbbb: I think it's pretty flexible, you can just put any string you want, and you can avoid supplying those values if you are manually inserting values for some reason.

diyor28: If you have defaults at the application layer as well, then it's ok

modules/core/migrations/20240602202521-initial.sql
diyor28: If you're concerned about storage efficiency, then don't. That's a user problem))

modules/core/migrations/20240602202521-initial.sql
diyor28: You're right. My knowledge of postgres is way outdated turns out)

modules/core/migrations/20240602202521-initial.sql
jcbbb: Changed to int

PR #21, opened 2024-11-26
IS-37 Логин - поправить респонс

jcbbb:


pkg/presentation/controllers/login_controller.go
```diff
@@ -1,16 +1,56 @@
...
+	"github.com/iota-agency/iota-sdk/pkg/service"
 	"github.com/iota-agency/iota-sdk/pkg/services"
 	"github.com/iota-agency/iota-sdk/pkg/types"
-	"net/http"
 
 	"github.com/gorilla/mux"
 	"github.com/iota-agency/iota-sdk/pkg/composables"
 )
 
+func SetFlash(w http.ResponseWriter, name string, value []byte) {
```
jcbbb: Where can I put these helper functions?

diyor28: Wait a second. You decided to embed errors in Cookies? Why not return them in html as we do with other pages?

pkg/presentation/controllers/login_controller.go
```diff
@@ -30,31 +70,61 @@ func (c *LoginController) Register(r *mux.Router) {
...
+		ErrorMessage: string(errorMessage),
+	}).Render(r.Context(), w); err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
 }
 
 func (c *LoginController) Post(w http.ResponseWriter, r *http.Request) {
-	if err := r.ParseForm(); err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+	dto := LoginDTO{
```
diyor28: It'd be better if you switched to using Decoder/Validator as we do in other controllers.

diyor28 (changes requested):

pkg/presentation/controllers/login_controller.go
diyor28: Why are you embedding errors in cookies in the first place?

jcbbb: So that I don't have to provide props for the pages again. They are called flash messages (don't know if you know this), and widely used in SSR apps.

diyor28: Oh, did not know that. I would put this in the same package as ParseId & Redirect functions

PR #22, opened 2024-11-27
IS-13 Save user form does not work

jcbbb:


pkg/infrastructure/persistence/mappers.go
```diff
@@ -35,7 +35,8 @@ func toDomainUser(dbUser *models.User) *user.User {
 		MiddleName: middleName,
 		Email:      dbUser.Email,
 		Password:   password,
-		//AvatarID:   dbUser.AvatarID,
+		Avatar:     toDomainUpload(dbUser.Avatar),
```
diyor28: I would rather move the nil check here. toDomainUpload should seg fault when handling a nil pointer

pkg/presentation/mappers/mappers.go
```diff
@@ -102,6 +108,9 @@ func EmployeeToViewModel(entity *employee.Employee) *viewmodels.Employee {
 }
 
 func UploadToViewModel(entity *upload.Upload) *viewmodels.Upload {
+	if entity == nil {
```
diyor28: Here as well, I think it's better to keep these functions seg faulting when handling nil pointers

diyor28 (changes requested):

pkg/presentation/mappers/mappers.go
jcbbb: Done

pkg/infrastructure/persistence/mappers.go
jcbbb: Done

PR #23, opened 2024-11-27
IS-43 Do not render pagination for total < limit

jcbbb:


components/base/pagination/pagination.templ
```diff
@@ -6,34 +6,36 @@ import (
...
-			} else {
-				<li>
-					<a
-						class={ "btn btn-secondary border-none rounded-none h-11 px-5", page.Classes() }
-						href={ templ.SafeURL(page.Link) }
-					>
-						{ fmt.Sprintf("%d", page.Num) }
-					</a>
-				</li>
+	if len(state.Pages()) > 0 {
```
diyor28: It's better to keep this in the parent components. Since it introduces an obscure behavior similar to that nil pointer situation in a different PR.

I can already imagine some new dev using this component and not understanding why nothing renders.
Yes, he can look at the source code and find this out, but let's be honest not many people are smart enough for this:)

diyor28 (changes requested):

components/base/pagination/pagination.templ
jcbbb: This is exactly what I thought when you had this code in the first place. I removed `len(state.Pages()) > 0`, now brought it back )

jcbbb: fixed

diyor28: Yes, my bad, mindlessly copied it from a different project

PR #24, opened 2024-11-27
IS-10 Positions image upload

diyor28:


modules/core/locales/en.json
```diff
@@ -245,6 +245,7 @@
         "Title": "Account"
       }
     },
+    "Avatar": "Browse filesystem",
```
jcbbb: This should probably be at the root, not inside `Account` object, and it's a noun describing an action, making it inconsistent with others. Maybe `BrowseFilesystem` would be better

jcbbb: @diyor28

diyor28: I'll rename it. The reason it's part of Account object is because it's a label passed from Account form to component

modules/warehouse/persistence/position_repository.go
```diff
@@ -89,17 +89,34 @@ func (g *GormPositionRepository) CreateOrUpdate(ctx context.Context, data *posit
...
 	if !ok {
 		return service.ErrNoTx
 	}
-	if err := tx.Create(toDBPosition(data)).Error; err != nil {
+	positionRow, junctionRows := toDBPosition(data)
+	if err := tx.Create(positionRow).Error; err != nil {
 		return err
 	}
+	for _, junctionRow := range junctionRows {
+		// TODO: this feels like a hack
```
diyor28, line 114: @jcbbb This feels a little off, not sure how to make it better

jcbbb: I think there is a way for gorm to do this with associations, no?

diyor28: Yes, but it would save uploads also. In my case uploads already exist. 
And also doing this in a for loop feels wrong too. Don't see any bulk action support in Gorm 

We may just switch to raw sql😂

jcbbb: > Don't see any bulk action support in Gorm

https://gorm.io/docs/create.html#Batch-Insert

> We may just switch to raw sql😂

The sooner we learn to start new projects with raw sql, the better :smile:. I would like to see your hatred towards ORMs develop more.

diyor28: I do dislike ORM's mostly. I initially started it without and ORM, but got tired of writing boiler plate, so switched to an ORM. But now thinking of going back

diyor28: That's great, I'll switch to using batch insert

PR #26, opened 2024-11-28
IS-29 Migrations should run as part of a single transaction

jcbbb:


jcbbb:
Removed `BEGIN` and `COMMIT` from migration files

PR #27, opened 2024-12-01
IS-36 SDK - страница настроек

jcbbb:


pkg/infrastructure/persistence/tab_repository.go
```diff
@@ -0,0 +1,106 @@
...
+func (g *GormTabRepository) GetAll(ctx context.Context, params *tab.FindParams) ([]*tab.Tab, error) {
+	tx, ok := composables.UseTx(ctx)
+	if !ok {
+		return nil, service.ErrNoTx
+	}
+	q, err := helpers.ApplySort(tx, params.SortBy)
+	if err != nil {
+		return nil, err
+	}
+	if params.UserID != 0 {
```
diyor28: Do not overload `GetAll` method with filtering by UserId. Instead create a separate `GetUserTabs` method

pkg/infrastructure/persistence/tab_repository.go
```diff
@@ -0,0 +1,106 @@
...
+		q.Where("user_id = ?", params.UserID)
+	}
+	var entities []*models.Tab
+	if err := q.Find(&entities).Error; err != nil {
+		return nil, err
+	}
+
+	tabs := make([]*tab.Tab, len(entities))
```
diyor28: Use `mapping.MapDbModels` (however you will have to change `ToDomainTab` return signature. 
Or use `mapping.MapViewModels`

pkg/infrastructure/persistence/tab_repository.go
```diff
@@ -0,0 +1,106 @@
...
+		return service.ErrNoTx
+	}
+	tab := ToDBTab(data)
+	if err := tx.Save(tab).Error; err != nil {
+		return err
+	}
+	return nil
+}
+
+func (g *GormTabRepository) Delete(ctx context.Context, params *tab.DeleteParams) error {
```
diyor28: Here as well

diyor28 (changes requested):

pkg/presentation/viewmodels/models.go
```diff
@@ -116,3 +116,9 @@ type Upload struct {
 	CreatedAt string
 	UpdatedAt string
 }
+
+type Tab struct {
+	ID       string
+	Href     string
+	Position uint
```
diyor28: This should be a string, unless there is a good reason

diyor28 (changes requested):

pkg/infrastructure/persistence/tab_repository.go
jcbbb: What's wrong with filtering?

pkg/presentation/viewmodels/models.go
jcbbb: Don't need this in viewmodel, removed it

pkg/infrastructure/persistence/tab_repository.go
diyor28: I was lazy to write it out myself, so here's the same thing from ChatGPT:

image

jcbbb: fixed

pkg/infrastructure/persistence/tab_repository.go
jcbbb: fixed

PR #38, opened 2024-12-20
resolve graphql loaders from single path

jcbbb:
All registered schemes should be able to resolve from same `/query`, but playground types are messed up. You would have to go to `/playground/{name}` to get correct types for testing for now.


diyor28:
Wow, expected it to be way easier

diyor28:
Please merge main into your branch and reflect the changes that took place since you created the branch

jcbbb:
Like this?

diyor28:
> image
> 
> I have to convert convert these to my own handlers?

Convert them to use the graphql registration approach you introduced

jcbbb:
@diyor28 I only left one graphql controller in core module, which is enough. All other modules should register their schema only.

PR #39, opened 2024-12-23
IS-49 Raw SQL PR

jcbbb:


jcbbb:
@diyor28 Check it out

diyor28:
@jcbbb looks good, but I expected you would rewrite on repo to showcase how the raw SQL approach would look

jcbbb:
> @jcbbb looks good, but I expected you would rewrite one of the repos to showcase how the raw SQL approach would look

I did, inventory_repository.go

modules/warehouse/persistence/inventory_repository.go
```diff
@@ -17,121 +19,400 @@ func NewInventoryRepository() inventory.Repository {
...
 	}
 
-	if params.Type != "" {
-		tx = tx.Where("type = ?", params.Type)
+	if params.Query != "" && params.Field != "" {
+		where, args = append(where, fmt.Sprintf("ic.%s::VARCHAR ILIKE $%d", params.Field, len(args)+1)), append(args, "%"+params.Query+"%")
 	}
 
-	tx, err := helpers.ApplySort(tx, params.SortBy)
+	rows, err := pool.Query(ctx, `
```
jcbbb, line 59: This query could become a problem if you are using graphql endpoints, and you want `inventory_checks->created_by->avatar`, for example. Maybe it's better to drop joins, and use individual selects?

```go
if check.CreatedBy, err = userRepo.FindUserByID(ctx, check.CreatedByID) {

}
```

diyor28: Yeah, why not. Most ORMs just don't do joins and do separate Selects. If it starts hitting performance at some point we'll come back to this

modules/core/infrastructure/persistence/mappers.go
```diff
@@ -16,11 +16,11 @@ import (
 	"github.com/iota-agency/iota-sdk/pkg/domain/entities/upload"
 )
 
-func ToDomainUser(dbUser *models.User) *user.User {
+func ToDomainUser(dbUser *models.User) (*user.User, error) {
 	roles := make([]*role.Role, len(dbUser.Roles))
-	for i, r := range dbUser.Roles {
-		roles[i] = toDomainRole(&r)
-	}
+	// for i, r := range dbUser.Roles {
```
diyor28, line 25: Did you meant to leave this?

jcbbb: No, but I will clean it up later when it's ready to merge

modules/core/infrastructure/persistence/authlog_repository.go
```diff
@@ -14,91 +23,124 @@ func NewAuthLogRepository() authlog.Repository {
...
 	}
-	return entities, nil
+	return logs, nil
 }
 
 func (g *GormAuthLogRepository) Count(ctx context.Context) (int64, error) {
-	tx, ok := composables.UseTx(ctx)
-	if !ok {
-		return 0, composables.ErrNoTx
+	pool, err := composables.UsePool(ctx)
```
diyor28, line 74: You should use `UsePoolTx`. In case `Count` gets called inside a transaction wrapped request, the user will expect changes done inside transaction to affect the call to `Count`

modules/core/infrastructure/persistence/authlog_repository.go
```diff
@@ -14,91 +23,124 @@ func NewAuthLogRepository() authlog.Repository {
...
-	if !ok {
-		return nil, composables.ErrNoTx
-	}
-	var entities []*authlog.AuthenticationLog
-	if err := tx.Find(&entities).Error; err != nil {
-		return nil, err
-	}
-	return entities, nil
+	return g.GetPaginated(ctx, &authlog.FindParams{
+		Limit: 100000,
```
diyor28, line 89: Why?

diyor28 (changes requested):

internal/server/default.go
```diff
@@ -26,6 +28,7 @@ func Default(options *DefaultOptions) (*server.HttpServer, error) {
 		middleware.Provide(constants.AppKey, app),
 		middleware.Provide(constants.HeadKey, layouts.Head()),
 		middleware.Provide(constants.LogoKey, layouts.DefaultLogo()),
+		middleware.Provide(constants.PoolKey, options.Pool),
 		middleware.Provide(constants.DBKey, options.Db),
```
diyor28, line 32: Make sure to remove gorm completely

PR #60, opened 2025-01-08
Update Dockerfile

Yagniksojitra:
Fixed docker file


Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN go install github.com/a-h/templ/cmd/templ@v0.2.793 && go install github.com/
 COPY go.mod go.sum ./
 RUN go mod download
 COPY . .
-RUN make generate && go vet ./...
+RUN make generate && go get ./...
```
diyor28, line 19: `go get` is not necessary since `go mod download` does the job

PR #94, opened 2025-01-17
feat: use a development container that supports hot reloading

abdelmounim-dev:
- The development container supports hot reloading.
- docs/CONTRIBUTING.MD includes comprehensive setup and usage documentation.
- Developers can run and test the application inside the container without additional setup.
- File changes trigger automatic reloads in the running container.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,35 @@
+FROM golang:1.23.2-alpine
+
+WORKDIR /app
+
+# Install required system dependencies
+RUN apk add --no-cache \
+    gcc \
+    musl-dev \
+    git \
+    nodejs \
+    npm \
+    curl \
+    make
+
+# Install Air for hot reloading
+RUN curl -sSfL https://raw.githubusercontent.com/cosmtrek/air/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.61.5
+
+# Install TailwindCSS
+RUN npm install -g tailwindcss@3.4.13
+
+# Install templ
+RUN go install github.com/a-h/templ/cmd/templ@v0.2.793
+
+# Copy go.mod and go.sum files
+COPY go.mod go.sum ./
+
+# Download Go modules
+RUN go get ./...
+
+# Copy the rest of the application
+COPY . .
+
+# Set Air to run on container startup
+ENTRYPOINT ["air"]
+
--- a/Makefile
+++ b/Makefile
@@ -35,9 +35,13 @@ coverage-score:
 report:
 	go tool cover -html=coverage.out -o ./coverage/cover.html
 
+# Dev containers
+dev-docker:
+	docker compose -f docker-compose.dev.yml up --build
+
 # Run PostgreSQL
 localdb:
-	docker compose -f docker-compose.dev.yml up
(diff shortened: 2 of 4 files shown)


docs/CONTRIBUTING.MD
diyor28, line 1: No need to override the local setup docs, better to just add a separate setup with docker section

docker-compose.dev.yml
```diff
@@ -13,3 +34,7 @@ services:
       - "5432:5432"
     volumes:
       - ./postgres-data:/var/lib/postgresql/data
+
+volumes:
+  go-modules:
```
diyor28: Why is this necessary? Doesn't docker just cache the installation step already?

Makefile
```diff
@@ -52,11 +56,11 @@ migrate-down:
 
 # Compile TailwindCSS (with watch)
 css-watch:
-	tailwindcss -c tailwind.config.js -i $(TAILWIND_INPUT) -o $(TAILWIND_OUTPUT) --minify --watch
+	npx tailwindcss -c tailwind.config.js -i $(TAILWIND_INPUT) -o $(TAILWIND_OUTPUT) --minify --watch
 
 # Compile TailwindCSS (without watch)
 css:
-	tailwindcss -c tailwind.config.js -i $(TAILWIND_INPUT) -o $(TAILWIND_OUTPUT) --minify
+	npx tailwindcss -c tailwind.config.js -i $(TAILWIND_INPUT) -o $(TAILWIND_OUTPUT) --minify
```
diyor28: Please remove these

Dockerfile.dev
```diff
@@ -0,0 +1,35 @@
...
+    nodejs \
+    npm \
+    curl \
+    make
+
+# Install Air for hot reloading
+RUN curl -sSfL https://raw.githubusercontent.com/cosmtrek/air/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.61.5
+
+# Install TailwindCSS
+RUN npm install -g tailwindcss@3.4.13
```
diyor28: Use installation instructions from the prod file. To install the standalone api without the need for nodejs

Dockerfile.dev
diyor28, line 1: Also I think you can turn this file into a separate target in the main file

diyor28 (changes requested):

docker-compose.dev.yml
abdelmounim-dev: it isn't necessary but it helps reduce the build time after adding or removing dependencies, if you add a dependency during development it wouldn't be added to layer cached by docker, so by having the dependencies in a volume they wouldn't need to be redownloaded after restarting the container.
but since we wouldn't be updating deps that often it wouldn't make that big of a difference

diyor28: I see, nice!

docs/CONTRIBUTING.MD
abdelmounim-dev: done

Makefile
abdelmounim-dev: done

Dockerfile.dev
abdelmounim-dev: done

PR #99, opened 2025-01-26
feat: add localization values to resolve custom linter findings

twinguy:
This PR resolves the missing localization key/value pairs identified by the custom linter that was recently introduced.  Most of the missing values were in the ru.json localization files.  I also added in the Uzbek (uz.json) translations but both the Russian and Uzbek translations need to have someone fluent in them to confirm.  

All translation values were generated using a GenAI LLM approach.


diyor28:
Thank you! I think LLM generated is ok for now. We'll fix them in bulk with a professional translator at some point

diyor28 (approved):

Issue #78, opened 2025-01-12
Z-index issues


diyor28:
## Reproduction
1. Go to `/warehouse/orders`
2. Click on the profile image

<img width="235" alt="image" src="https://github.com/user-attachments/assets/d7c09eae-51b5-4104-b167-50ff7e11ce80" />

PR #105, opened 2025-02-04
fix: z-index on `DetailsDropdown` component call-sites

thelissimus-work:
Fixes the issue: https://github.com/iota-uz/iota-sdk/issues/78

The main issue was the stacking context of the parent. Because both dropdowns used the same component (`DetailsDropdown`) there wasn't really a way to control the z-index from the call-site and browser's defaults kicked-in.

Tested on Safari and Firefox.

Aftermath of changes:

Screenshot 2025-02-04 at 20 10 29

### DIFF
--- a/components/base/dropdown.templ
+++ b/components/base/dropdown.templ
@@ -18,10 +18,24 @@ templ DropdownItem(props DropdownItemProps) {
 	</li>
 }
 
-templ DetailsDropdown(summary templ.Component, rootClass string) {
-	<div class={rootClass}>
+type DetailsDropdownProps struct {
+	Summary templ.Component
+	Classes templ.CSSClasses
+}
+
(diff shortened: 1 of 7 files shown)


components/base/dropdown.templ
```diff
@@ -18,8 +18,8 @@ templ DropdownItem(props DropdownItemProps) {
 	</li>
 }
 
-templ DetailsDropdown(summary templ.Component) {
```
diyor28, line 21: Handle this using a `Props` struct. And also use `templ.CSSClasses` type for passing in css classes, gives more flexibility to the end user

diyor28:
Also I do not see changes to the `templ` version

components/base/dropdown.templ
thelissimus-work: @diyor28 Done. Updated the docs (`COMPONENTS.md`) while I was at it too.

PR #107, opened 2025-02-05
refactor(modules): extract `bichat` from `core` into its own module

thelissimus-work:
Moments that I am still unsure of are:

- `bichat.templ` still referencing core template layout
- `chatfuncs` services were moved to `bichat` (I couldn't find a module that referenced it, so I moved it just in case because it seemed more relevant to `bichat` than to `core`)
- `modules/bichat/module.go` the `Register` function. Please verify if I correctly initialized the necessary services/controllers
- `dialogue_events.go` still references `core` domain user aggregate.

Overall, I assumed it is okay that modules can reference `core` when necessary (such as user aggregate) but not vice versa to prevent cyclic dependencies.


diyor28:
> Moments that I am still unsure of are:
> 
> * `bichat.templ` still referencing core template layout
> * `chatfuncs` services were moved to `bichat` (I couldn't find a module that referenced it, so I moved it just in case because it seemed more relevant to `bichat` than to `core`)
> * `modules/bichat/module.go` the `Register` function. Please verify if I correctly initialized the necessary services/controllers
> * `dialogue_events.go` still references `core` domain user aggregate.
> 
> Overall, I assumed it is okay that modules can reference `core` when necessary (such as user aggregate) but not vice versa to prevent cyclic dependencies.

You are absolutely correct in your assumptions. Referencing down the chain is okay and inevitable. Vise versa not acceptible

PR #135, opened 2025-02-18
fix(eventbus): log warning when no matching subscribers for an event

thelissimus-work:
> [!NOTE]
> - Used existing `logging` (which uses `logrus` under the hood) module. Not sure though. Happy to comply and rewrite to use just the `logrus` itself.
> - Added tests too. Seemed kinda redundant, but I think it is worth it just to at least play around with the isolated module itself instead of starting the whole app.

Example logs:

```json
{"level":"warning","msg":"eventbus.Publish: no matching subscribers for event with args: [\u003c*moneyaccount.CreatedEvent Value\u003e]","time":"2025-02-18T13:30:50+05:00"}
```

### DIFF
--- a/pkg/commands/migrate_command.go
+++ b/pkg/commands/migrate_command.go
@@ -5,16 +5,15 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"runtime/debug"
 	"time"
 
 	"github.com/iota-uz/iota-sdk/modules"
 	"github.com/iota-uz/iota-sdk/pkg/application"
 	"github.com/iota-uz/iota-sdk/pkg/configuration"
 	"github.com/iota-uz/iota-sdk/pkg/eventbus"
-	"github.com/iota-uz/iota-sdk/pkg/logging"
 
 	"github.com/jackc/pgx/v5/pgxpool"
-	"github.com/sirupsen/logrus"
 )
 
 var (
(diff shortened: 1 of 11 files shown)


diyor28:
> We can also consider the option of putting logs in `MatchSignature` function too with debug level. There are much information to be logged if needed.

Don't see a point

pkg/commands/migrate_command.go
```diff
@@ -31,7 +34,7 @@ func Migrate(mods ...application.Module) error {
 	if err != nil {
 		panic(err)
 	}
-	app := application.New(pool, eventbus.NewEventPublisher())
+	app := application.New(pool, eventbus.NewEventPublisher(logging.ConsoleLogger(logrus.WarnLevel)))
```
diyor28: This feels wrong. A logger should be a singleton that is defined once

diyor28 (changes requested):

pkg/commands/migrate_command.go
thelissimus-work: It is **defined** once though, in `logger.go`. Here I am just instantiating a logger with that definition + the right logging level for the current module/executable/test. But I am happy to comply and make it a global variable if you want inside of that very `logger.go` with:

```go
var ConsoleWarnLogger = ConsoleLogger(logrus.WarnLevel)
```

Issue #139, opened 2025-02-18
[Test] Cypress E2E Test for Editing a User


diyor28:
## Description
Create a **Cypress end-to-end test** to verify that **editing an existing user** works as expected. This test should cover the full flow from **logging in**, **navigating to the users table**, **editing a user's details**, and **confirming the changes** in the table.

---

## Objectives

1. **Login Flow**:
   - Use the `cy.session()` API to handle user authentication efficiently.
   - Login with a test user account.

2. **Navigate to Users Table**:
   - Visit `/users` and confirm the user list is visible.

3. **Edit an Existing User**:
   - Locate an existing user in the table.
   - Click on **Edit** (e.g., a link or button).
   - Modify fields such as:
     - First Name
     - Last Name
     - Email
     - Role or other relevant fields
   - Submit the changes.

4. **Verify Changes**:
   - Return to the users table.
   - Validate that the changes to the user’s details are **reflected in the table**.

5. **Post-Edit Login Validation (Optional)**:
   - If the **email or password was changed**, attempt to **log in with updated credentials** to confirm changes are effective.

---

## Suggested Test Flow (Cypress)
Example test structure outline:
```js
describe("User edit flow", () => {
  beforeEach(() => {
    login("test@gmail.com", "TestPass123!");
  });

  afterEach(() => {
    cy.visit("http://localhost:3200/logout");
  });

  it("edits a user and verifies changes in users table", () => {
    cy.visit("http://localhost:3200/users");

    // Locate the user to edit (e.g., by email or row position)
    cy.get("tbody tr")
      .contains("test1@gmail.com")
      .parent()
      .find('a[href*="/users/edit"]')
      .click();

    // Edit user fields
    cy.get("[name=FirstName]").clear().type("UpdatedName");
    cy.get("[name=LastName]").clear().type("UpdatedLast");
    cy.get("[name=Email]").clear().type("updated.email@example.com");
    // Update other fields as necessary...

    cy.get("[id=save-btn]").click();

    // Verify changes in the table
    cy.visit("http://localhost:3200/users");
    cy.get("tbody tr").contains("updated.email@example.com").should("exist");
  });
});

PR #147, opened 2025-02-21
feat(e2e): editing a user

thelissimus-work:
Closes: #139

Currently working on resetting the database.


Makefile
```diff
@@ -37,11 +37,16 @@ report:
 
 # Run PostgreSQL
 localdb:
-	docker compose -f compose.dev.yml up
+	docker compose -f compose.dev.yml up -d
```
diyor28: You shouldn't really do this, since this is meant to be used for local development. To keep in foreground and be able to shut it down using ctrl + c

e2e/cypress/e2e/users/register.cy.js
```diff
@@ -43,4 +45,28 @@ describe("user auth and registration flow", () => {
 		cy.url().should("include", "/users");
 		cy.get("tbody tr").should("have.length", 2);
 	});
+
+	it("edits a user and displays changes in users table", () => {
+		login("test1@gmail.com", "TestPass123!");
```
diyor28, line 50: Isn't it `test@gmail.com`?

thelissimus-work: That's the admin though. I'm editing a newly created user.

diyor28: Doesn't the DB reset after each test though?

thelissimus-work: It resets after each `describe`, not `it`.

thelissimus-work: Oh, in fact, before, not after. Each suite starts from a clean sheet of paper so to speak.

diyor28: Got it

Issue #145, opened 2025-02-21
[Bug] User's middle name is not displayed on edit


thelissimus-work:
<img width="1423" alt="Image" src="https://github.com/user-attachments/assets/568dfe00-a3e4-48b8-99fd-9a7cec7a6c66" />

Middle name is not displayed even though it is explicitly input in Cypress test:

https://github.com/iota-uz/iota-sdk/blob/4a722d74c01da7802178dbd2d9fcb7675ef5e064/e2e/cypress/e2e/users/register.cy.js#L25

Issue #146, opened 2025-02-21
[Bug] User delete modal buttons lack text


thelissimus-work:
<img width="398" alt="Image" src="https://github.com/user-attachments/assets/0379e799-a480-44be-b6eb-94fa1ba5acb5" />

PR #152, opened 2025-02-24
fix: user create and update Middle name

thelissimus-work:
Closes: #145 

Have to update the tests when #147 is merged too.


diyor28:
> To be fair, Go's zero-values were a mistake. It just swallows incomplete struct initialization without any (type) error. This was the cause of the issue #146 too. We have to rely on an external analyzer because of this.
> 
> Having said that, why did we use `exhaustruct` only with mappers? Why not enable it throughout the whole project? This could've been easily caught by it.
> 
> https://github.com/iota-uz/iota-sdk/blob/be6cc4974df511a3633278655a71780916290e58/.golangci.yml#L7-L16

Cuz most structs do not need this behaviour, also I couldn't set up `exhaustruct` & `golangci-lint` properly. At the moment it's not enable in any of the CI pipelines

PR #153, opened 2025-02-25
feat(schema): add column rename detection and drop table syntax

twinguy:
# Change Summary

1. **Enhanced Schema Diff Generator:**
   - Added column rename detection functionality that intelligently identifies when columns are being renamed rather than just dropped and added
   - Implemented DROP TABLE syntax support for schema migrations
   - Added DROP COLUMN syntax support
   - Added logic to track tables and columns that have been dropped in previous migrations to avoid redundant operations

2. **Migration Management:**
   - Added functionality to generate proper down migration statements for renamed columns and dropped tables
   - Implemented a system to check existing migrations to prevent conflicts

3. **Logger Initialization Fixes:**
   - Fixed a bug in the schema collector where the logger wasn't being properly initialized with default log levels
   - Added a default log level (InfoLevel) when none is specified
   - Properly initialized the logger in the Generator struct

4. **Code Refactoring:**
   - Replaced global logger references with instance-specific logger references (g.logger instead of logger)
   - Improved error handling and logging consistency


twinguy:
@diyor28 - I have completed quite a bit of additional testing on the schema change detector and it feels like it is operating as expected.  I did notice I failed to include the DROP TABLE and default logging changes in the last PR so I have added them into this PR for review.  Also would be interested in hearing any other testing results from others when available.

twinguy:
Looks good to me.

Issue #185, opened 2025-03-11
Account balance can't equal zero


alisherKhadimetov:
Actual: 

<img width="1437" alt="Image" src="https://github.com/user-attachments/assets/00709cd5-5d4a-4e62-93cc-6a49d1c27bce" />
While creating an account, account balance can't be zero

Expected:
Account balance CAN be zero.

PR #193, opened 2025-03-11
fix: 0 balance value

thelissimus-work:
`required` and `gte=0` contradict each other. `required` checks if not zero.

Also, why use `float64` for a balance? Are we really sure it is not going to create problems down the road?

Closes: #185


diyor28:
> `required` and `gte=0` contradict each other. `required` checks if not zero.
> 
> Also, why use `float64` for a balance? Are we really sure it is not going to create problems down the road?
> 
> Closes: #185

It shouldn't be float64, I did it this way in a rush

Issue #196, opened 2025-03-12
When updating client info passport data is inserted instead of updated


diyor28:
![Image](https://github.com/user-attachments/assets/8c5b8fab-acad-44f3-8bb8-c4a1e9d95daa)

PR #199, opened 2025-03-12
fix: update instead of insert on client edit

thelissimus-work:
Closes: #196


modules/core/domain/entities/passport/passport.go
```diff
@@ -133,6 +133,19 @@ func New(series, number string, opts ...Option) Passport {
 	return p
 }
 
+func NewWithID(id uuid.UUID, series, number string, opts ...Option) Passport {
```
diyor28: No, no. Add a `WithID` option instead

diyor28 (changes requested):

Issue #188, opened 2025-03-11
Can't create an employee


alisherKhadimetov:
<img width="1710" alt="Image" src="https://github.com/user-attachments/assets/8fc04390-d93a-4bf7-8a24-5b1e809bdfc6" />

Server returns 500

alisherKhadimetov:
@thelissimus-work still it doesnt work, even with your recommendation. 

<img width="1702" alt="Image" src="https://github.com/user-attachments/assets/1c798d8c-da7e-4591-9313-f846c7d46b1c" />

<img width="1709" alt="Image" src="https://github.com/user-attachments/assets/1add993a-0fb7-45ea-b0a9-8094f36b64d9" />

PR #204, opened 2025-03-13
fix: show error messages on invalid PIN/TIN and etc

thelissimus-work:
Closes: #188

### DIFF
--- a/modules/hrm/presentation/controllers/employee_controller.go
+++ b/modules/hrm/presentation/controllers/employee_controller.go
@@ -9,7 +9,6 @@ import (
 	"github.com/a-h/templ"
 	"github.com/go-faster/errors"
 	"github.com/gorilla/mux"
-	"github.com/nicksnyder/go-i18n/v2/i18n"
 
 	"github.com/iota-uz/iota-sdk/modules/core/domain/entities/currency"
 	"github.com/iota-uz/iota-sdk/modules/core/domain/value_objects/money"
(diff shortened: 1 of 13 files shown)


modules/hrm/presentation/controllers/employee_controller.go
```diff
@@ -166,8 +168,62 @@ func (c *EmployeeController) Create(w http.ResponseWriter, r *http.Request) {
...
+		errorsMap := map[string]string{}
+		if errors.Is(err, tax.ErrInvalidTin) {
+			tinFieldName := l.MustLocalize(&i18n.LocalizeConfig{
+				MessageID: "Employees.Private.TIN.Label",
+			})
+			errorsMap["Tin"] = l.MustLocalize(&i18n.LocalizeConfig{
+				MessageID: "ValidationErrors.custom",
+				TemplateData: map[string]string{
+					"Field": tinFieldName,
+					"Error": err.Error(),
```
diyor28: You're not localizing the error itself though. I think the custom key is redundant. The field name is obvious on the frontend. Instead you should focus on translating the error

diyor28 (dismissed):

modules/hrm/presentation/controllers/employee_controller.go
diyor28: Also look into `pkg/serrors`. It was meant to be a universal error handling solution to simplify this process. Some work on the package maybe needed

Issue #169, opened 2025-03-07
Menu items in sidebar are not set by default


alisherKhadimetov:
Expected: All available items in the navigation sidebar are visible.

Actual: Sidebar navigation at the first login is empty.

alisherKhadimetov:
@thelissimus-work 

<img width="1710" alt="Image" src="https://github.com/user-attachments/assets/3f542b74-b1ef-4d29-9362-ba799a76b298" /> 
New user

diyor28:
@sabokhat1 This should be already fixed in production

PR #211, opened 2025-03-18
fix: permission issues and sidebar item defaults

thelissimus-work:
Closes: #169


modules/core/presentation/controllers/user_controller.go
```diff
@@ -252,6 +263,14 @@ func (c *UsersController) GetEdit(w http.ResponseWriter, r *http.Request) {
 }
 
 func (c *UsersController) GetNew(w http.ResponseWriter, r *http.Request) {
+	if err := composables.CanUser(r.Context(), permissions.UserCreate); err != nil {
```
diyor28: Move permission checks to services. You will be duplicating this logic when implementing graphql resolvers. Also change `CanUser` signature to be:
`(context.Context, ....permission.Permission)` this way you can inline these two checks.
Even better if you can provide a few utility functions so that SDK user can do:

```go
composables.CanUser(ctx, rbac.OR(permissions.UserRead, permisisons.RoleRead))
```

```go
composables.CanUser(ctx, rbac.OR(permissions.ViewBlog, rbac.AND(permissions.UserRead, permisisons.RoleRead)))
```

diyor28 (changes requested):

pkg/rbac/rbac.go
```diff
@@ -0,0 +1,62 @@
+package rbac
+
+import (
+	"github.com/iota-uz/iota-sdk/modules/core/domain/aggregates/user"
+	"github.com/iota-uz/iota-sdk/modules/core/domain/entities/permission"
+)
+
+type Permission interface {
```
diyor28, line 16: Since you already created this package, please move the RBAC manager from `core/domain/entities/permissions` here

diyor28 (dismissed):

PR #215, opened 2025-03-24
added radio card, progress bar, datepicker

jcbbb:


components/base/input/datepicker.templ
```diff
@@ -0,0 +1,92 @@
...
+	Name         string
+	EndName      string
+	StartName    string
+	Form         string
+}
+
+templ DatePicker(props DatePickerProps) {
+	{{ pageCtx := composables.UsePageCtx(ctx) }}
+	{{ selected, _ := templ.JSONString(props.Selected) }}
+	<div x-data={ fmt.Sprintf("datePicker({locale: '%s', mode: '%s', dateFormat: '%s', labelFormat: '%s', minDate: '%s', maxDate: '%s', selectorType: '%s', selected: %s})", pageCtx.Locale.String(), props.Mode, props.DateFormat, props.LabelFormat, props.MinDate, props.MaxDate, props.SelectorType, selected) }>
```
diyor28: Won't this be easier to do with `templ` JSON calling? Like it's done in charts/charts.templ?

jcbbb: How? `fmt.Sprintf("datePicker(%s)", templ.JSONString(props))`?

diyor28: Right, you're not defining the function the same way I do in charts. Event the option you proposed looks better, don't u agree?

jcbbb: Check @diyor28

PR #222, opened 2025-03-24
feat: direct permissions to user

thelissimus-work:
I have some uncommited changed locally too.

### DIFF
--- a/modules/core/module.go
+++ b/modules/core/module.go
@@ -52,7 +52,7 @@ func (m *Module) Register(app application.Application) error {
 
 	app.RegisterServices(
 		services.NewUploadService(uploadRepo, fsStorage, app.EventPublisher()),
-		userService,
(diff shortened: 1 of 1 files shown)


modules/core/domain/aggregates/user/user.go
```diff
@@ -172,24 +183,27 @@ func New(
...
+	roles       []role.Role
+	groupIDs    []uuid.UUID
+	permissions []*permission.Permission
+	lastLogin   time.Time
+	lastAction  time.Time
+	createdAt   time.Time
+	updatedAt   time.Time
+}
+
+var _ User = (*user)(nil)
```
diyor28: I believe this check is not necessary since conformity to interface is checked in the user constructor

diyor28 (dismissed):

PR #228, opened 2025-03-25
don't require phone number in client constructor

jcbbb:
Since phone number is no longer required, I removed it from client constructor and added WithPhone option


jcbbb:
@diyor28

PR #231, opened 2025-03-27
Multitenancy

thelissimus-work:


pkg/middleware/auth.go
```diff
@@ -89,6 +89,15 @@ func ProvideUser() mux.MiddlewareFunc {
 					return
 				}
 				ctx = context.WithValue(ctx, constants.UserKey, u)
+
+				tenantService := app.Service(services.TenantService{}).(*services.TenantService)
+				t, err := tenantService.GetByID(ctx, u.TenantID())
+				if err != nil {
```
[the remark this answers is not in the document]
diyor28: @thelissimus-work This sounds reasonable

diyor28:
@thelissimus-work Use UUID for tenants table. We will be migrating to UUID pretty soon, so why not start using UUID for new tables already

modules/core/infrastructure/persistence/upload_repository.go
```diff
@@ -126,8 +137,14 @@ func (g *GormUploadRepository) Count(ctx context.Context) (int64, error) {
 	if err != nil {
 		return 0, err
 	}
+
+	tenant, err := composables.UseTenant(ctx)
```
mirolimsgd, line 136: I would suggest to require tenantId as an explicit parameter in function signatures where it is mandatory, instead of implicitly relying on its presence within a context object (ctx).
- Error Prevention: If tenantId is mandatory for the function's logic, requiring it in the signature prevents runtime errors that occur when it's missing from the ctx. This shifts error detection earlier in the development cycle.
- Reduced Boilerplate: This approach eliminates the repetitive code currently needed in many methods and repositories to extract and validate the tenantId from the context.
-  Improved Clarity & Explicitness: Making tenantId an explicit parameter clearly communicates the function's dependency, making the code easier to understand and maintain, thereby reducing the potential for runtime errors.

Handling Multi-Tenant vs. All-Tenant Queries:
For scenarios where a method needs to retrieve records across all tenants (i.e., be tenant-independent), we could handle this by passing a designated value for the tenantId parameter (e.g., an empty string, null, or a specific constant). The method implementation would then check this parameter's value to decide whether or not to include the tenant filtering condition in its database query.

bektosh-iota: That is indeed correct. I pointed this out to Diyor as well, for some reason tenant id propagation is done like this throughout the whole repo.
I'm reworking this whole thing

bektosh-iota: @mirolimsgd Regarding this issue.
It's been quite some time since this PR is hanging and because it's massive, the longer it halts, the more problems like merge conflicts and it's getting hard maintaining this PR. So maybe I could rework tenant id propagation in a different PR. I'd create a separate issue for it.
It's just that it's taking some time to do this. Besides, speeding up this with AI turned out to be a bit of problem, as it turns out AI still struggles with big refactorings across many files.

diyor28: @bektosh-iota This wasn't implemented by me, so implement it as you see fit.

modules/core/infrastructure/persistence/session_repository.go
```diff
@@ -25,16 +27,17 @@ const (
...
-        VALUES ($1, $2, $3, $4, $5, $6)`
+        VALUES ($1, $2, $3, $4, $5, $6, $7)`
 	sessionUpdateQuery = `
         UPDATE sessions
         SET expires_at = $1,
             ip = $2,
             user_agent = $3
-        WHERE token = $4`
-	sessionDeleteQuery = `DELETE FROM sessions WHERE token = $1`
+        WHERE token = $4 AND tenant_id = $5`
```
diyor28, line 39: Why is this necessary? Token is already a primary key

modules/core/infrastructure/persistence/session_repository.go
```diff
@@ -25,16 +27,17 @@ const (
...
+	sessionDeleteQuery = `DELETE FROM sessions WHERE token = $1 AND tenant_id = $2`
```
diyor28: Here as well

modules/core/infrastructure/persistence/session_repository.go
```diff
@@ -81,19 +91,53 @@ func (g *GormSessionRepository) Count(ctx context.Context) (int64, error) {
...
 	}
+
+	tenant, err := composables.UseTenant(ctx)
+	if err != nil {
+		return 0, err
+	}
+
 	var count int64
-	if err := tx.QueryRow(ctx, sessionCountQuery).Scan(&count); err != nil {
+	if err := tx.QueryRow(ctx, sessionCountQuery+" WHERE tenant_id = $1", tenant.ID).Scan(&count); err != nil {
```
diyor28: For constructing sql queries in the runtime use methods from the `repo` package. Reference other usages of this package to get an idea on how to use it

modules/core/infrastructure/persistence/upload_repository.go
diyor28: I see the problem now. Tenant is already part of domain entities and db models, so passing it through context makes no sense

modules/core/domain/entities/tenant/tenant.go
```diff
@@ -0,0 +1,86 @@
+package tenant
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type Tenant struct {
```
diyor28, line 9: Is there a particular reason Tenant isn't an interface like other domain entities?

modules/core/infrastructure/persistence/setup_test.go
```diff
@@ -33,23 +34,45 @@ func setupTest(t *testing.T) *testFixtures {
...
 	tx, err := pool.Begin(ctx)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	t.Cleanup(func() {
-		if err := tx.Commit(ctx); err != nil {
-			t.Fatal(err)
+		// Rollback instead of commit to ensure clean state
+		// This is safer as it ensures tests don't affect each other
```
diyor28, line 63: They already don't affect each other since each test runs in a separate db. Committing changes is useful to be able to introspect the db state after a failed test

modules/core/infrastructure/persistence/upload_repository.go
mirolimsgd: > @mirolimsgd Regarding this issue. It's been quite some time since this PR is hanging and because it's massive, the longer it halts, the more problems like merge conflicts and it's getting hard maintaining this PR. So maybe I could rework tenant id propagation in a different PR. I'd create a separate issue for it. It's just that it's taking some time to do this. Besides, speeding up this with AI turned out to be a bit of problem, as it turns out AI still struggles with big refactorings across many files.

It is just suggestions. I am not blocking pr. The problem I see if it is merged it then it will be used and if not immediately fixed will create more work later so I suggest to change it now. It will reduce amount of code in this pr and later there is no need to spend more time.
Both options work for me.

pkg/composables/tenant.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/google/uuid"
+	"github.com/iota-uz/iota-sdk/pkg/constants"
+)
+
+var (
+	ErrNoTenantFound = errors.New("no tenant found in context")
+)
+
+type Tenant struct {
```
diyor28, line 15: What is this duplicate type for?

modules/core/infrastructure/persistence/upload_repository.go
diyor28: > > @mirolimsgd Regarding this issue. It's been quite some time since this PR is hanging and because it's massive, the longer it halts, the more problems like merge conflicts and it's getting hard maintaining this PR. So maybe I could rework tenant id propagation in a different PR. I'd create a separate issue for it. It's just that it's taking some time to do this. Besides, speeding up this with AI turned out to be a bit of problem, as it turns out AI still struggles with big refactorings across many files.
> 
> It is just suggestions. I am not blocking pr. The problem I see if it is merged it then it will be used and if not immediately fixed will create more work later so I suggest to change it now. It will reduce amount of code in this pr and later there is no need to spend more time. Both options work for me.

I agree

PR #239, opened 2025-04-01
breadcrumb component

jcbbb:


components/base/breadcrumb/breadcrumb.templ
```diff
@@ -0,0 +1,44 @@
+// Example usage:
```
diyor28, line 3: I think you need to prefix the comment with package name, otherwise it won't be recognized by go doc command

jcbbb: had no idea it was for go doc, I thought it was just a comment :)

diyor28: Uhu, It is for human and LLM usage

PR #258, opened 2025-04-16
fix: 243 tx to InTx for account,tab,user

danil-bragin:


modules/core/presentation/controllers/account_controller.go
```diff
@@ -55,7 +55,7 @@ func (c *AccountController) Register(r *mux.Router) {
 
 	setRouter := r.PathPrefix(c.basePath).Subrouter()
 	setRouter.Use(commonMiddleware...)
-	setRouter.Use(middleware.WithTransaction())
+	//setRouter.Use(middleware.WithTransaction())
```
diyor28: You can completely remove this line

modules/core/services/user_service.go
```diff
@@ -65,41 +65,39 @@ func (s *UserService) GetPaginatedWithTotal(ctx context.Context, params *user.Fi
...
-
-	if err := tx.Commit(ctx); err != nil {
-		return err
-	}
-	createdEvent.Result = created
+	createdEvent.Result = createdUser
 	s.publisher.Publish(createdEvent)
-	return nil
+
+	return err
```
diyor28, line 100: You're checking for `err != nil` on line 94. Returning `nil` is the correct behaviour

modules/core/services/user_service.go
```diff
@@ -65,41 +65,39 @@ func (s *UserService) GetPaginatedWithTotal(ctx context.Context, params *user.Fi
...
 	if err != nil {
 		return err
 	}
-	defer func() {
-		if err := tx.Rollback(ctx); err != nil {
-			logger.WithError(err).Error("failed to rollback transaction")
+
+	var createdUser user.User
+	err = composables.InTx(ctx, func(txCtx context.Context) error {
+		data, err = data.SetPassword(data.Password())
```
diyor28: This doesn't necessarily need to be part the transaction

modules/core/handlers/tab_handler.go
```diff
@@ -59,12 +60,20 @@ func (h *TabHandler) createUserTabs(ctx context.Context, user user.User) {
 
 	if len(tabs) > 0 {
 		ctxWithUser := context.WithValue(ctx, constants.UserKey, user)
-		if err := h.tabService.DeleteUserTabs(ctxWithUser, user.ID()); err != nil {
-			h.logger.Errorf("Failed to delete existing tabs for user %d: %v", user.ID(), err)
-			return
-		}
-		if err := h.tabService.CreateMany(ctxWithUser, tabs); err != nil {
-			h.logger.Errorf("Failed to create tab for user %d: %v", user.ID(), err)
+		err := composables.InTx(ctx, func(txCtx context.Context) error {
```
diyor28: This call is not necessary since you're opening transactions inside service calls. If you want both calls to be part of a single transaction, create a service method that wraps these two operations in a single transaction

diyor28:
@danil-bragin

modules/core/presentation/controllers/account_controller.go
the-npden4ik: fixed

modules/core/services/user_service.go
the-npden4ik: fixed

modules/core/handlers/tab_handler.go
the-npden4ik: made by calling the service

Issue #265, opened 2025-04-18
Raw SQL Error Exposed on Duplicate Email During User Creation


the-npden4ik:
**User Create/Update: Required field validations not working (similar to phone validation issue)**

This issue is similar to the previously reported phone validation bug: https://github.com/iota-uz/iota-sdk/issues/242

On both the **User Create** and **User Update** pages, validation rules (e.g., `required`) are not working as expected. When submitting an empty or incomplete form, no validation errors are shown to the user. This leads to confusion, as there’s no indication of which fields are missing or invalid.

Please verify that:
- Backend validation is correctly triggered for required fields.
- Validation errors are properly passed to the frontend.
- The frontend correctly renders those errors.

PR #266, opened 2025-04-18
fix: #265 impl create/update user validation & show error messages

danil-bragin:


modules/core/infrastructure/persistence/user_repository.go
```diff
@@ -248,6 +248,9 @@ func (g *PgUserRepository) GetByEmail(ctx context.Context, email string) (user.U
 	if len(users) == 0 {
 		return nil, errors.Wrap(ErrUserNotFound, fmt.Sprintf("email: %s", email))
 	}
+	if len(users) == 0 {
```
diyor28: This code is unreachable

danil-bragin: fixed by this [commit](https://github.com/iota-uz/iota-sdk/pull/266/commits/38c37fb5595e73a7c83d7acaa2e6a26b064789dc)

Issue #330, opened 2025-07-02
Bug: crud-controller — Fields with readonly: true should be disabled (non-editable) on the update (edit) form


danil-bragin:
In the crud-controller, fields marked with the property readonly: true are currently editable in the update (edit) form. This violates expected behavior, as such fields should be rendered as disabled inputs — visible, but not editable.

islom-d-iota:
@claude Provide me an implementation plan

PR #344, opened 2025-07-04
fix: crud_controller — non-editable fields with readonly: true

islom-d-iota:


islom-d-iota:
fixes #330

danil-bragin:
@islom-d-iota merged

PR #386, opened 2025-07-16
docs: consolidate and refine js-runtime specification

diyor28:


diyor28:
@claude Turn docs/js-runtime markdown files into proper github issues

PR #424, opened 2025-07-23
fix: updated column sorting logic, introduced ui

islom-d-iota:


islom-d-iota:
@danil-bragin @diyor28 

правки по сортировке колонок в scaffolding#table

components/scaffold/table/helpers.go
```diff
@@ -106,6 +106,15 @@ func WithSortURL(sortURL string) ColumnOpt {
 	}
 }
 
+// WithNotSortable disables sorting for a column
+func WithNotSortable() ColumnOpt {
```
diyor28: This is a little weird, why not make column not sortable by default and add Sortable as needed?

PR #434, opened 2025-07-29
kanban_board: generic kanban board (initial impl);

jcbbb:


danil-bragin:
@claude do review again

Issue #373, opened 2025-07-15
Checkbox in edit group is not clickable


diyor28:
<img width="461" height="1365" alt="Image" src="https://github.com/user-attachments/assets/6c86b539-600d-4dac-b4fa-dba42b10b253" />

This is a low priority task @jcbbb

PR #450, opened 2025-08-06
fix: fix id collision in checkboxes in groups

jcbbb:
Fixed id collision


jcbbb:
https://github.com/iota-uz/iota-sdk/issues/373

PR #652, opened 2026-03-02
feat(core): add S3 upload backend and presigned download flow

diyor28:
## Summary
- extend upload storage contract with delete and presigned-get URL support
- add S3 storage backend implementation with MinIO-compatible endpoint/path-style config
- wire runtime backend selection via `UPLOAD_STORAGE_BACKEND=fs|s3`
- make upload download serving backend-aware while preserving existing URL path compatibility
- update upload API response URL generation and delete semantics to use storage backend
- add service tests for presigned URL + physical delete behavior

## Config
- `UPLOAD_STORAGE_BACKEND`
- `UPLOAD_S3_BUCKET`, `UPLOAD_S3_REGION`, `UPLOAD_S3_ENDPOINT`
- `UPLOAD_S3_ACCESS_KEY`, `UPLOAD_S3_SECRET_KEY`
- `UPLOAD_S3_SECURE`, `UPLOAD_S3_FORCE_PATH_STYLE`
- `UPLOAD_S3_PRESIGN_TTL_SECONDS`

## Validation
- `go test ./modules/core/infrastructure/persistence ./modules/core/services ./modules/core/presentation/controllers -count=1`

## Summary by CodeRabbit

* **New Features**
  * Added S3 storage backend option for uploads (filesystem remains default)
  * Implemented presigned URL generation for secure, time-limited download links
  * Added file deletion capability for uploaded content
  * Enhanced upload API responses with direct download URLs

* **Configuration**
  * New S3 settings available: storage backend selection, bucket, region, credentials, endpoint, and presigned link TTL

### DIFF
--- a/modules/core/infrastructure/persistence/upload_s3_storage.go
+++ b/modules/core/infrastructure/persistence/upload_s3_storage.go
@@ -1,0 +1,160 @@
+package persistence
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/signer/v4"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/iota-uz/iota-sdk/pkg/configuration"
+)
+
+type s3Client interface {
+	PutObject(ctx context.Context, params *s3.PutObjectInput, optFns ...func(*s3.Options)) (*s3.PutObjectOutput, error)
+	GetObject(ctx context.Context, params *s3.GetObjectInput, optFns ...func(*s3.Options)) (*s3.GetObjectOutput, error)
+	CopyObject(ctx context.Context, params *s3.CopyObjectInput, optFns ...func(*s3.Options)) (*s3.CopyObjectOutput, error)
+	DeleteObject(ctx context.Context, params *s3.DeleteObjectInput, optFns ...func(*s3.Options)) (*s3.DeleteObjectOutput, error)
+}
+
+type s3PresignClient interface {
+	PresignGetObject(ctx context.Context, params *s3.GetObjectInput, optFns ...func(*s3.PresignOptions)) (*v4.PresignedHTTPRequest, error)
+}
+
+type S3Storage struct {
+	bucket  string
+	client  s3Client
+	presign s3PresignClient
+}
+
+func NewS3Storage() (*S3Storage, error) {
+	conf := configuration.Use()
+
+	bucket := strings.TrimSpace(conf.UploadS3Bucket)
+	region := strings.TrimSpace(conf.UploadS3Region)
+	endpoint := normalizeS3Endpoint(strings.TrimSpace(conf.UploadS3Endpoint), conf.UploadS3Secure)
+	accessKey := strings.TrimSpace(conf.UploadS3AccessKey)
+	secretKey := strings.TrimSpace(conf.UploadS3SecretKey)
+
+	if bucket == "" {
+		return nil, fmt.Errorf("UPLOAD_S3_BUCKET is required")
+	}
+	if region == "" {
+		region = "us-east-1"
+	}
+	if accessKey == "" || secretKey == "" {
+		return nil, fmt.Errorf("UPLOAD_S3_ACCESS_KEY and UPLOAD_S3_SECRET_KEY are required")
+	}
+
+	awsCfg, err := awsconfig.LoadDefaultConfig(
+		context.Background(),
+		awsconfig.WithRegion(region),
+		awsconfig.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(accessKey, secretKey, "")),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("load aws config: %w", err)
+	}
+
+	client := s3.NewFromConfig(awsCfg, func(o *s3.Options) {
+		o.UsePathStyle = conf.UploadS3ForcePathStyle
+		if endpoint != "" {
+			o.BaseEndpoint = aws.String(endpoint)
+		}
+	})
+
+	return &S3Storage{
+		bucket:  bucket,
+		client:  client,
+		presign: s3.NewPresignClient(client),
+	}, nil
+}
+
+func (s *S3Storage) Open(ctx context.Context, fileName string) ([]byte, error) {
+	key := normalizeS3Key(fileName)
+	res, err := s.client.GetObject(ctx, &s3.GetObjectInput{
+		Bucket: aws.String(s.bucket),
+		Key:    aws.String(key),
+	})
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = res.Body.Close() }()
+
+	return io.ReadAll(res.Body)
+}
+
+func (s *S3Storage) Save(ctx context.Context, fileName string, b []byte) error {
+	key := normalizeS3Key(fileName)
+	_, err := s.client.PutObject(ctx, &s3.PutObjectInput{
+		Bucket: aws.String(s.bucket),
+		Key:    aws.String(key),
+		Body:   bytes.NewReader(b),
+	})
+	return err
+}
+
+func (s *S3Storage) Rename(ctx context.Context, oldPath, newPath string) error {
+	sourceKey := normalizeS3Key(oldPath)
+	targetKey := normalizeS3Key(newPath)
+
+	_, err := s.client.CopyObject(ctx, &s3.CopyObjectInput{
+		Bucket:     aws.String(s.bucket),
+		CopySource: aws.String(s.bucket + "/" + sourceKey),
+		Key:        aws.String(targetKey),
+	})
+	if err != nil {
+		return err
+	}
+	return s.Delete(ctx, sourceKey)
+}
+
+func (s *S3Storage) Delete(ctx context.Context, fileName string) error {
+	key := normalizeS3Key(fileName)
+	_, err := s.client.DeleteObject(ctx, &s3.DeleteObjectInput{
+		Bucket: aws.String(s.bucket),
+		Key:    aws.String(key),
+	})
+	return err
+}
+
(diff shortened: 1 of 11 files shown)


modules/core/services/upload_service_test.go
```diff
@@ -0,0 +1,257 @@
...
+			path: "uploads/a.txt",
+			setup: func(storage *uploadStorageMock) {
+				storage.On("PresignGetURL", mock.Anything, "uploads/a.txt", mock.Anything).Return("", errors.New("presign failed")).Once()
+			},
+			expectedError: true,
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
```
diyor28, line 162: **🔴 Critical (process) — this `tc := tc` is the actual cause of the red "Code Quality & Formatting" check (it's a lint failure, not gofmt).**

`just check lint` (golangci-lint) fails with `The copy of the 'for' variable "tc" can be deleted (Go 1.22+) (copyloopvar)` at this line **and `:235`**, exit 1. Go 1.22+ gives each iteration a fresh `tc`, so `tc := tc` is dead and the linter rejects it. (For the record: `gofmt -l`/`gofmt -s -l` are clean on all 11 files — the "formatting" hypothesis is wrong; it's this lint rule.) Notably the immediately-preceding repo commit `4a43a8996 "fix(tests): satisfy copyloopvar…"` fixed this exact rule elsewhere — it was reintroduced here.

**Fix:** delete both `tc := tc` lines.

modules/core/infrastructure/persistence/upload_s3_storage.go
```diff
@@ -0,0 +1,203 @@
...
+	return body, nil
+}
+
+func (s *S3Storage) Save(ctx context.Context, fileName string, b []byte) error {
+	const op serrors.Op = "persistence.S3Storage.Save"
+	key, err := normalizeS3Key(fileName)
+	if err != nil {
+		return serrors.E(op, err)
+	}
+	_, err = s.client.PutObject(ctx, &s3.PutObjectInput{
```
diyor28, line 110: **🟠 Major (correctness) — `PutObject` never sets `ContentType`, so every object lands as `application/octet-stream` and presigned GETs force a download instead of inline render.**

This breaks the whole point of the serve path: ``/PDF previews won't render — the browser downloads them. The upload entity already has a detected mimetype (`mimetype.Detect` in `upload_dto.go`), but `Storage.Save([]byte)` discards it — the interface has no way to pass content type through.

**Fix:** extend `Save` (or add a typed variant) to carry content type and set `ContentType` on `PutObjectInput`; optionally set `ResponseContentType`/`ResponseContentDisposition` on the presign request. (Both CodeRabbit and Copilot missed this.)

modules/core/services/upload_service.go
```diff
@@ -141,11 +144,42 @@ func (s *UploadService) CreateMany(ctx context.Context, data []*upload.CreateDTO
...
+	return downloadURL, nil
+}
+
 func (s *UploadService) Delete(ctx context.Context, id uint) (upload.Upload, error) {
 	entity, err := s.repo.GetByID(ctx, id)
 	if err != nil {
 		return nil, err
 	}
+	if err := s.storage.Delete(ctx, entity.Path()); err != nil {
+		return nil, err
```
diyor28, line 181: **🟠 Major (correctness) — the delete reorder just moved the partial-failure window; both returns are also raw (no `serrors`).**

The earlier Copilot note (DB-then-storage) is now stale — this was reordered to storage-first. But the non-atomicity just inverted: if `storage.Delete` succeeds and `repo.Delete` then fails, you keep a **DB row pointing at an object that no longer exists** (broken downloads), with no rollback. Storage-first is arguably the lesser evil (delete is idempotent on both FS and S3), but it's still unhandled. Separately, both `return nil, err` paths are raw, violating the `serrors.Op`/`serrors.E` convention these same files otherwise follow.

**Fix:** wrap both with a `services.UploadService.Delete` op, and decide ordering deliberately — the usual pattern is DB-first inside a tx, then best-effort storage delete with failures logged for a reclaim job.

modules/core/module.go
```diff
@@ -271,3 +274,15 @@ func uploadAPIControllerOpts(opts *ModuleOptions) []controllers.UploadAPIControl
...
+	conf := configuration.Use()
+	switch strings.ToLower(strings.TrimSpace(conf.UploadStorageBackend)) {
+	case "", "fs":
+		return persistence.NewFSStorage()
+	case "s3":
+		return persistence.NewS3Storage()
+	default:
+		return nil, fmt.Errorf("unsupported upload storage backend: %s", conf.UploadStorageBackend)
+	}
+}
```
diyor28, line 288: **🟠 Major (sloppy / convention) — raw `fmt.Errorf` on the storage-backend startup path, inconsistent with the rest of this code.**

`newUploadStorage` returns `fmt.Errorf("unsupported upload storage backend: %s", …)` while its siblings `NewFSStorage`/`NewS3Storage` both define a `serrors.Op` and wrap. This is a boot-config path that should fail fast with a typed error.

**Fix:** `const op serrors.Op = "core.Module.newUploadStorage"` and `return nil, serrors.E(op, serrors.Invalid, …)`. (Credit where due: the fail-fast behavior itself is correct — `NewS3Storage` rejects missing bucket/keys, so boot aborts on bad S3 config.)

modules/core/presentation/controllers/upload_controller.go
```diff
@@ -138,3 +144,49 @@ func (c *UploadController) Create(w http.ResponseWriter, r *http.Request) {
...
+
+	basePath := strings.Trim(path.Clean("/"+conf.UploadsPath), "/")
+	storagePath := path.Join(basePath, normalized)
+	if !strings.HasPrefix(storagePath, basePath+"/") {
+		http.NotFound(w, r)
+		return
+	}
+
+	downloadURL, err := c.uploadService.GetDownloadURLByPath(r.Context(), storagePath)
+	if err != nil {
```
diyor28, line 187: **🟡 Minor (correctness) — collapsing every error into `404` makes an S3 outage look identical to a missing file.**

Any error from presign generation (S3 unreachable, bad credentials, throttling) becomes `http.NotFound`. That defeats monitoring and misleads clients during an infra incident.

**Fix:** distinguish not-found (`serrors.NotExist` → 404) from infrastructure failure (→ 502/500), and log the underlying error.

modules/core/infrastructure/persistence/upload_s3_storage.go
```diff
@@ -0,0 +1,203 @@
...
+		return serrors.E(op, err)
+	}
+	targetKey, err := normalizeS3Key(newPath)
+	if err != nil {
+		return serrors.E(op, err)
+	}
+
+	_, err = s.client.CopyObject(ctx, &s3.CopyObjectInput{
+		Bucket:     aws.String(s.bucket),
+		CopySource: aws.String(s.bucket + "/" + sourceKey),
```
diyor28, line 134: **🟡 Minor (correctness) — `CopySource` isn't URL-encoded, and `Rename` is a non-atomic copy-then-delete.**

`CopySource: aws.String(s.bucket + "/" + sourceKey)` must be URL-encoded for keys with spaces/unicode (AWS SDK v2 won't do it for you here). Keys are currently ASCII hex-md5 slugs, so this is latent rather than live — but it'll silently corrupt copies if key generation ever changes. Also, if `DeleteObject` fails after `CopyObject` (`:132–140`), both objects persist with no rollback.

**Fix:** `url.PathEscape` each key segment in `CopySource`; document/handle the non-atomicity.

diyor28 (commented):
## 🤖 Codex PR review — `feat(core): add S3 upload backend and presigned download flow`

Read-only review (gofmt clean; `go vet` passes; verified the S3 client is constructed once and reused, MinIO works via `UPLOAD_S3_ENDPOINT` + force-path-style, and config validation fails fast). **6 inline comments above.** This is SDK core consumed by EAI (which will switch to MinIO via config), so correctness + the config surface matter.

### 🔴 Blocker
- **The red CI check is a `copyloopvar` lint, not gofmt** — `tc := tc` at `upload_service_test.go:162` and `:235` (reintroduced after a prior repo commit fixed the same rule). See inline.

### 🟠 Notable (inline)
- **No `ContentType` on `PutObject`** → presigned GETs serve `application/octet-stream`, breaking inline image/PDF rendering (`upload_s3_storage.go:110`). Both bots missed this.
- **Delete reorder moved the partial-failure window** (now: object gone, DB row dangling) + raw errors (`upload_service.go:181`).
- **Raw `fmt.Errorf` on the backend-select startup path** (`module.go:288`).

### Additional findings
- **Missing config docs (`.env.example`).** The PR adds 9 config fields (`UPLOAD_STORAGE_BACKEND`, `UPLOAD_S3_BUCKET/REGION/ENDPOINT/ACCESS_KEY/SECRET_KEY/SECURE/FORCE_PATH_STYLE/PRESIGN_TTL_SECONDS` — `environment.go:248`) but doesn't add them to `.env.example`, the repo's config-doc surface. Downstream EAI/MinIO operators can't discover them, and the secret vars have no placeholder. Add all nine with safe defaults + a one-line backend note.
- **Minor:** `upload_api_controller.go:191` uses stdlib `log.Printf` (new `"log"` import) instead of the request-scoped structured logger — loses correlation/fields. (The fallback-to-`entity.URL()` behavior itself is fine and resolves the earlier "presign failure orphans the upload" concern — that's now non-fatal.)
- **Minor (tests):** the only new controller test exercises the traversal-reject branch (returns 404 before touching the nil service). The happy path (valid key → 303 + presigned `Location`) is untested and would nil-panic with the empty-struct construction; there's no test that the S3 backend actually routes through `ServeUploadByPath`.
- **Nit:** `_ = ctx` / `_ = ttl` dead-assignments in `upload_fs_storage.go` (Go doesn't error on unused *params* — just drop them); and the FS `PresignGetURL` returns a self-referential absolute URL that would loop if FS were ever routed through `ServeUploadByPath` (harmless today).

### ✅ Already addressed by `55792b8d6` — not re-flagging (prior bot comments are stale)
- dir mode `0777` → `0o755` ✅ · `normalizeS3Key` now errors on blank/dot input ✅ · presign-failure no longer orphans the upload (logs + falls back) ✅ · path traversal in `ServeUploadByPath` is well-mitigated (mux pre-clean + explicit `..` reject + final prefix check) — residual is the test gap above, not an exploit ✅
- Also note: the Copilot claim that the FS URL "regressed from absolute to relative" is **incorrect** — it's absolute (`Scheme+Domain+Path`), same as the old `entity.URL()`.

Issue #639, opened 2026-03-01
Lint debt follow-up: re-enable deferred staticcheck style rules and clean violations


diyor28:
## Context
To unblock CI for PR #637, we fixed actionable lint errors and temporarily deferred broad style-only staticcheck debt by excluding these checks in `.golangci.yml`:

- `ST1000` (package comments)
- `ST1003` (naming/acronym style)
- `ST1016` (receiver name consistency)
- `ST1020` (exported func/method comment format)
- `ST1021` (exported type comment format)
- `ST1022` (exported var comment format)

This issue tracks paying that debt down and re-enabling the checks.

## Why this matters
- Keeps codebase conventions consistent
- Prevents drift in naming/docs style
- Reduces future churn when stricter linting is restored

## Scope
1. Audit all current violations for the excluded ST checks.
2. Batch fixes by package/module (small, reviewable PRs).
3. Prefer safe renames with backward-compatible aliases when APIs are exported and externally consumed.
4. Update affected tests/docs/call sites.
5. Re-enable checks in `.golangci.yml` incrementally (one or two at a time).

## Suggested rollout
1. Re-enable `ST1020`, `ST1021`, `ST1022` first (comment-only, low risk).
2. Re-enable `ST1016` next (receiver consistency, low-medium risk).
3. Re-enable `ST1000` (package comments) across modules.
4. Re-enable `ST1003` last (naming; highest churn/risk).

## Acceptance criteria
- All excluded ST checks are re-enabled.
- `just check lint` passes in CI without temporary style exclusions.
- Any exported renames include migration notes in PR descriptions/changelog where applicable.

## References
- PR: https://github.com/iota-uz/iota-sdk/pull/637
- Lint config: `.golangci.yml`

diyor28:
Started a focused slice of this lint-debt cleanup in [PR #752](https://github.com/iota-uz/iota-sdk/pull/752): `just check lint` now runs a dedicated staticcheck pass for `ST1000`, `ST1016`, `ST1020`, `ST1021`, and `ST1022` on handwritten Go while skipping generated files (`*_templ.go`, `*_gen.go`, `generated.go`, and files marked `Code generated` / `DO NOT EDIT`). Repo-wide batching showed `ST1016` and `ST1020`-`ST1022` are clean under that boundary; `ST1000` noise is currently dominated by generated/template-derived files, so this moves enforcement forward without mixing generated-file churn into the handwritten-code cleanup.

PR #752, opened 2026-04-15
[codex] complete issue 715 cleanup and start #714 follow-up refactors

diyor28:
## Summary
- remove the dead deprecated compatibility surfaces that were still left in-tree for issue #715
- move group realtime subscriptions out of controller registration into composition-managed event handlers
- reduce a small but noisy chunk of startup duplication by centralizing the shared process wrapper and IOTA runtime construction helpers in `pkg/bootstrap`

## Why
Issue #715 tracks the remaining compatibility leftovers after the composition refactor. A few of the listed items were already gone in the current tree, but several deprecated symbols and legacy fallbacks were still present and no longer had in-repo callers.

This PR also folds in focused slices of two broader cleanup issues where the changes were already close at hand:
- #714: event subscriptions should not be registered from presentation controllers; config/global access should be trimmed where constructor injection is already available; process bootstrap code should not be copy-pasted across entrypoints
- #639: investigation landed, but the extra lint wrapper was intentionally removed to keep the tooling simple

The OIDC storage case from #715 remains slightly subtle: `GetSigningKey` is no longer part of the current `github.com/zitadel/oidc/v3` `op.Storage` contract, so it could be removed safely. `SetUserinfoFromScopes` is still required by that upstream interface, so this PR keeps it and removes the misleading deprecation marker instead of deleting the method.

Closes #715
Partially addresses #714

## Developer Impact
- `authlog.AuthenticationLog` compatibility struct is gone; use `authlog.AuthLog`
- the legacy `NewGotenbergPDFExporter(...)` path and its fallback behavior are removed in favor of `NewExportToPDFTool(...)`
- `money.Amount` is no longer exported; use `int64` or `Money.BigAmount()` depending on range needs
- group realtime websocket updates now follow the same composition-managed event-handler pattern as user realtime updates instead of subscribing from controller registration
- user/group realtime broadcasters now take an injected logger instead of reaching into the global config singleton for logging
- `cmd/server`, `cmd/worker`, and `cmd/superadmin` now share the same bootstrap panic/error wrapper and runtime-construction helper

## Validation
- `go vet ./...`
- `go test ./pkg/money ./pkg/bichat/tools/export`
- `go test ./modules/oidc/infrastructure/oidc -run TestStorageKeySet_ExposesOnlyPublicKeys`
- `go test ./pkg/bootstrap ./cmd/server ./cmd/worker ./cmd/superadmin`
- `go test ./modules/core/...`
- `just check lint`


diyor28:
Superseded by and rolled into #753. Closing; branch will be deleted.

PR #781, opened 2026-05-26
feat(core): organizational model — departments + user positions

diyor28:
Implements the backend foundation of #780 — a reusable organizational-unit model in `core`, the predecessor for EAI's EDO department-scoped access control (iota-uz/eai#2865).

## What

**Aggregates**
- `department` — hierarchical (self-referencing `parent_id`), tenant-scoped, multilingual `name` (jsonb), `order`/`status`.
- `userposition` — user↔department link with `is_manager` (dept-head) and `is_primary` flags, multilingual `title`.

**Read service** — `OrgQuery` (consumed by downstream access resolvers):
- `UserDepartments(userID)` — departments the user holds a position in
- `UserManagedDepartments(userID, includeSubtree)` — departments a user heads; subtree walk includes descendants
- `DepartmentSubtree(deptID)` — department + all descendants

Hierarchy walks use **recursive CTEs**, tenant-scoped on both anchor and recursive terms. **No Postgres views or triggers.**

**Write validation** (shared by services *and* seeders): 4-locale completeness (en/ru/uz/uz-Cyrl), cross-tenant rejection for `parent_id` / `department_id` / `user_id`, self-reference and re-parent **cycle** guards — all `serrors.KindValidation`.

**Plus**: `Department`/`Position` CRUD permissions (seeded), seed plumbing, DI wiring, migration `changes-1779808179.sql` (+ `core-schema.sql` sync).

## Design notes
- Mirrors the existing `group` aggregate vertical slice (interfaces, empty repos + `composables.UseTx`, `pkg/repo` builders, immutable setters).
- `user_positions.user_id` is `integer` to match `users.id`; all other PKs/tenant ids are `uuid`.
- Tenant integrity is enforced **app-side** (not composite FKs): a self-referencing `parent_id` with `ON DELETE SET NULL` is incompatible with a composite `(id, tenant_id)` FK, and app-level validation keeps `users`/schema untouched.

## Verification
- `go vet ./modules/core/...` and `./modules/... ./cmd/...` — clean
- `gofmt -l modules/core/` — clean
- Full new suite against a real Postgres — **49 pass / 0 fail**, incl. cross-tenant isolation (subtree never bleeds across tenants), structural/cross-tenant validation rejections, and re-parent cycle rejection
- Migration up → down → up — clean

## Follow-ups (not in this PR)
- Admin CRUD UI for departments + positions
- Real EAI org-chart seed data (plumbing is in place via `DepartmentsSeedFunc`/`UserPositionsSeedFunc`)

[...]

PR #783, opened 2026-05-27
feat(core): tenant-qualified unique on user_groups (tenant_id, id)

diyor28:
## What
Adds `UNIQUE (tenant_id, id)` to `user_groups` (migration `changes-1779900000.sql` + schema snapshot).

## Why
Downstream modules need to declare tenant-qualified composite foreign keys against `user_groups`:
```sql
FOREIGN KEY (tenant_id, group_id) REFERENCES user_groups (tenant_id, id)
```
so a row can never reference a group in another tenant — enforced by the database, not only the application. `id` is already the primary key, so the composite is always satisfiable for existing rows (no data change, safe additive constraint).

Mirrors the `users_tenant_id_id_key` / `departments_tenant_id_id_key` constraints added in the core org-model PR (#781).

## Unblocks
Tenant-qualified participant FKs in the EAI EDO granular-permissions work (iota-uz/eai#2865), addressing a Codex review finding (cross-tenant participant rows possible at the DB level).

## Risk
Minimal — single additive UNIQUE constraint; no data migration.

## Summary by CodeRabbit

* **Chores**
  * Enhanced database schema with improved data integrity constraints.

[Review Change Stack](https://app.coderabbit.ai/change-stack/iota-uz/iota-sdk/pull/783?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/migrations/changes-1779900000.sql
+++ b/migrations/changes-1779900000.sql
@@ -1,0 +1,14 @@
+-- Migration: Tenant-qualified unique key on user_groups
+-- Purpose: Add a UNIQUE (tenant_id, id) constraint to user_groups so downstream
+--   modules can declare tenant-qualified composite foreign keys
+--   (e.g. FOREIGN KEY (tenant_id, group_id) REFERENCES user_groups (tenant_id, id)),
+--   preventing cross-tenant references at the database level. id is already the
+--   primary key, so the composite is always satisfiable for existing rows.
+
+-- +migrate Up
+ALTER TABLE user_groups
+    ADD CONSTRAINT user_groups_tenant_id_id_key UNIQUE (tenant_id, id);
+
+-- +migrate Down
+ALTER TABLE user_groups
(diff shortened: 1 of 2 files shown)


diyor28:
Re: the `user_groups` `name` uniqueness comment — that's a **pre-existing** constraint (global `name UNIQUE` from the original table + `UNIQUE (tenant_id, name)`), not introduced by this PR. This change only **adds** `UNIQUE (tenant_id, id)` to enable tenant-qualified composite FKs; it doesn't touch the name uniqueness.

The observation is fair (the global `name` unique makes the tenant-scoped one redundant and blocks per-tenant name reuse), but dropping a long-standing unique constraint is a behavioral, data-sensitive change that warrants its own deliberate change rather than a drive-by in this narrow infra PR. Resolving here; can be tracked separately if per-tenant group names are desired.

PR #789, opened 2026-05-28
fix(core/departments): unblock dept-scoped visibility + drawer validation UX

diyor28:
Three follow-ups from live E2E testing of [iota-uz/eai#2893](https://github.com/iota-uz/eai/pull/2893) (EDO granular permissions). All three live in `iota-sdk` because the affected surface is the org-model + admin-CRUD layer landed via #781/#785. eai#2893 just needs the resulting submodule bump.

## 1. BLOCKER — drop OrgQuery's `DepartmentRead` gate
`OrgQueryService.UserDepartments/UserManagedDepartments/DepartmentSubtree` gated every call on `composables.CanUser(ctx, permissions.DepartmentRead)` in the **calling user's context**. Downstream resolvers (EDO `AccessResolver` in eai#2893) consume OrgQuery in user-context to build visibility scopes — so any role lacking `DepartmentRead` silently returned `ErrForbidden`, the resolver degraded to an empty dept set, and dept-based document visibility broke for **every non-admin user**. Admin testing alone didn't catch it (admin gets `view:all` short-circuit).

Confirmed live during eai#2893 E2E: granting `DepartmentRead` to the test user immediately restored subtree visibility.

The fix removes the three service-level perm checks. Safe because:
- `PgOrgQueryRepository` is already tenant-scoped on every query (no cross-tenant leak).
- The admin Department CRUD keeps its `DepartmentRead` gate at the controller layer — unchanged.
- Only callers in either repo are EDO `AccessResolver` and the Department admin controller; nothing else relies on the service-level gate.

Regression test `org_query_service_test.go::TestOrgQueryService_NoPermissionRequired` asserts an unprivileged actor can call all three methods without error.

## 2. Validation-class errors → field-level drawer feedback (no more silent 500s)
The Department admin controller returned `http.Error(w, "Error …ing department", 500)` for every service/repo error. The service already typed validation errors with `serrors.KindValidation`; the controller never read `Kind`. The repository didn't classify PG `23505`. Net effect: duplicate-code, self-loop, cycle, and invalid-UUID parent ID all surfaced as silent 500s with the drawer just sitting there.

**End-to-end shape:**
- Service errors become **typed sentinels** (`services.ErrDepartmentSelfLoop` / `ErrDepartmentCycle` / `ErrDepartmentParentNotFound`), wrapped via `fmt.Errorf("%w", …)` inside the existing `serrors.E` chain so `errors.Is` matches downstream.
- Repository classifies PG `SQLSTATE 23505` on `departments_tenant_id_code_key` via a new `classifyDepartmentDBError(op, err)` helper (`pgconn.PgError` + `ConstraintName` match), mirroring the precedent in `pkg/bichat/tools/sql/sql_error_classifier.go`. The constraint sentinel is `persistence.ErrDepartmentDuplicateCode`.
- Controller adds `departmentValidationFieldError(ctx, err) → (status, fieldErrs)` mapping each sentinel to its drawer field + HTTP status:

  | Sentinel | Field | Status | i18n key |
  |---|---|---|---|
  | `ErrDepartmentSelfLoop` | `ParentID` | 400 | `Departments.Errors.SelfLoop` |
  | `ErrDepartmentCycle` | `ParentID` | 400 | `Departments.Errors.Cycle` |
  | `ErrDepartmentParentNotFound` | `ParentID` | 400 | `Departments.Errors.ParentNotFound` |
  | `ErrDepartmentDuplicateCode` | `Code` | 409 | `Departments.Errors.DuplicateCode` |

- `Create` / `Update` error paths first try the helper; if it returns a field map, rebuild the same props the DTO-error branch uses and re-render the drawer with the right status header. Unknown errors still fall through to the existing 5xx path (catch-all intentionally not added — the drawer template renders per-field errors only, so a generic "_form" entry would silently disappear).

i18n: 4 new keys added across all 5 locale files (en/ru/uz/uz-Cyrl/zh).

## 3. Dedup parent dropdown
`ParentSelect` passed both `Placeholder:` to `base.Select` **and** rendered `<option value="">{NoParent}</option>` inside the slot — yielding two "Без родителя" entries (one disabled with no value, one clickable with `value=""`). Picking the wrong one posted the literal Russian text as `ParentID`, which then triggered bug 2's silent 500.

Kept the explicit clickable empty option (so "no parent" remains re-selectable after picking one) and dropped the `Placeholder:` arg.

## Verification
- `go vet ./...` ✅
- `go test -count=1 ./modules/core/services/... ./modules/core/infrastructure/persistence/... ./modules/core/infrastructure/query/... ./modules/core/presentation/controllers/...` → all green (services 7s, persistence 12s, query 18s, controllers 56s).
- `just check tr` ✅ 1791 keys consistent across 5 locales.
- `golangci-lint run --new` on touched paths → 0 issues.
- Manual E2E walkthrough during eai#2893 testing confirmed the bugs; this PR is the upstream fix that lets eai#2893 ship without a permissions workaround.

## Risks
- No callers outside this repo rely on the OrgQuery permission gate behavior (only EDO via eai consumes it; no SDK-side tests assert `forbidden` on these methods).
- `pgconn` is already vendored transitively via `pgx/v5` — no new dependency.
- `shared_templ.go` regenerated alongside the `.templ` edit.

## Follow-ups in eai
- Bump `back/go.mod` iota-sdk pseudo-version + `sdk/` submodule pointer in eai#2893.
- Optionally bump `back/modules/edo/services/access_resolver.go` log level from `.Warn` to `.Error` on the (now-rare) OrgQuery-error degradation path.
- No EAI seed change needed: `DepartmentRead` is no longer required on non-admin roles.

cc @iota-uz

## Summary by CodeRabbit

* **New Features**
  * Stronger department validation: prevents self-parenting and cycles and surfaces precise field-level errors with appropriate HTTP statuses.

* **Localization**
  * Added localized validation messages (en/ru/uz-Cyrl/uz/zh) for hierarchy errors and duplicate department codes.

* **Bug Fixes**
  * Fixed parent-select rendering to avoid duplicate options and improved form feedback for validation/conflict cases.

* **Behavior Change**
  * Department read queries no longer require an extra permission check in this internal API.

* **Tests**
  * Added tests for validation sentinels, duplicate-code handling, controller error mapping, and relaxed read-query behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/iota-uz/iota-sdk/pull/789?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


diyor28:
Addressed the AI-review feedback in [`abc6ce2`](https://github.com/iota-uz/iota-sdk/pull/789/commits/abc6ce226). Summary:

| # | Reviewer(s) | Issue | Fix |
|---|---|---|---|
| 1 | copilot + coderabbitai | `validateDepartmentParent` wrapped *every* `repo.GetByID` failure as `ErrDepartmentParentNotFound`, so transient DB/tx errors would render as a 400 "parent not found" field error in the drawer. | Only wrap when the underlying error `errors.Is(err, persistence.ErrDepartmentNotFound)`. Everything else propagates as a non-validation `serrors.E` so the controller renders 5xx. |
| 2 | copilot | Stale doc comment on `departmentValidationFieldError` claimed UUID-parse errors were mapped to a `ParentID` field error, but the switch doesn't include that case. | Updated the doc to reflect reality: malformed UUIDs are caught earlier by the DTO `omitempty,uuid` tag and surface through the existing DTO-error branch. |
| 3 | copilot | `classifyDepartmentDBError` doc said "callers should pass every Save/Update/Delete error through this helper," but `Delete()` doesn't use it. | Tightened the comment to describe the actual scope (create/update unique-constraint classification). Added a forward-looking note that a FK-violation classifier can be added when the admin cascade-delete UX is built. |
| 4 | coderabbitai | Sentinel test used 4 `t.Run` blocks; project guideline is table-driven. | Refactored `TestDepartmentValidationSentinels` to a single table with `{name, setup, action, expected}` entries. Added `t.Helper()` in the per-case setup hook for the `thelper` lint. |

All gates re-passed: `go vet ./...`, `golangci-lint run --new` (0 issues), `just check tr` (1791 keys × 5 locales), `go test -count=1` on services + persistence + query + presentation/controllers (all green).

diyor28:
Addressed the new @coderabbitai feedback in [`21f281a`](https://github.com/iota-uz/iota-sdk/pull/789/commits/21f281a8d). 

**Issue:** `services/org_validation.go` was depending on `infrastructure/persistence` directly to `errors.Is` against `persistence.ErrDepartmentNotFound` — coupling the service layer to a concrete infra package, against the SDK guideline "Keep domain and services decoupled from infrastructure details."

**Fix:** Moved the Repository contract's surface errors to the domain package where the `Repository` interface lives:

```go
// modules/core/domain/aggregates/department/department_repository.go
var (
    ErrNotFound      = errors.New("department not found")
    ErrDuplicateCode = errors.New("department: code already exists in tenant")
)
```

Persistence keeps `ErrDepartmentNotFound` / `ErrDepartmentDuplicateCode` as `= department.ErrXxx` aliases for backward-compat (existing tests + downstream EAI modules don't need to change). New call sites use the canonical domain sentinels:

- `services/org_validation.go` → `errors.Is(err, department.ErrNotFound)`; persistence import removed entirely.
- `controllers/department_controller.go::departmentValidationFieldError` → `errors.Is(err, department.ErrDuplicateCode)`.
- Both test files updated to reference the domain sentinel for clarity.

Repository implementations are responsible for wrapping the underlying driver error so `errors.Is(err, department.ErrNotFound)` resolves through the chain — `PgDepartmentRepository` already does this via `serrors.E` + the new `classifyDepartmentDBError` helper.

Gates: `go vet ./...`, `golangci-lint run --new` (0 issues), full core test suite — all green.

PR #793, opened 2026-06-02
[codex] Add declarative sidebar workspaces

### LINKED ISSUE #791: Sidebar navigation overhaul: admin-nav decoupling, stable keys, declarative workspaces, pinned items
**Scope:** overhaul of the SDK sidebar navigation subsystem. Four related parts (A–D) in the same `pkg/types` + `pkg/sidebar` + `pkg/middleware` + `components/sidebar` area. A & B unblock the EAI host today; C & D remove the structural footguns for every host. Downstream EAI re-wiring: iota-uz/eai#2941. EAI hotfix already shipped: iota-uz/eai#2940.

---

# Part A — Decouple core admin nav contribution from controller registration

## Problem

`core.NewComponent` unconditionally contributes `BuildNavItems()` = `[DashboardLink, AdministrationLink]` (`modules/core/component.go`). The only opt-out, `ModuleOptions.SkipAdminControllers`, gates **both** the nav contribution **and** the admin controllers (`/users`, `/roles`, `/groups`, `/departments`, `/positions`, `/settings/logo`, settings, sessions, spotlight, websocket).

A host that wants the admin **pages** but not the sidebar **group** (its own curated nav) has no clean option — it must let core inject the group and remove it downstream by inspecting children. There is no stable identity on `NavigationItem` to match against, and `Name` is localized before host-side filtering runs.

### Real-world breakage

EAI suppressed the Administration group by matching an exact 4-child count + href list + order. iota-sdk#785 (departments + positions admin CRUD) grew `AdministrationLink` to 6 children → the host match silently failed → the group reappeared in production (2nd such regression).

## Change

Add a dedicated option in `modules/core/component.go` `ModuleOptions` gating **only** the nav contribution:

```go
// SkipAdminNavItems suppresses contribution of the built-in admin
// navigation (dashboard + administration group) WITHOUT affecting
// controller registration. Use when the host provides its own curated
// sidebar but still wants the admin pages reachable.
SkipAdminNavItems bool
```

Gate the `composition.AddNavItems(builder, BuildNavItems(...))` call on this flag independently of `SkipAdminControllers`. Default `false` → current behavior preserved (non-breaking).

**Acceptance:**
- `SkipAdminNavItems: true` + `SkipAdminControllers: false` → admin pages and spotlight quick links still registered; no Dashboard/Administration nav contributed.
- All existing flag combinations unchanged when unset.

---

# Part B — Stable identity on `NavigationItem`

Add a stable, locale-independent identifier so hosts reference SDK nav declaratively instead of matching on content:

```go
type NavigationItem struct {
    Key  string // stable id, e.g. "core.administration"; locale-independent
    Name string // localized label
    ...
}
```

- Set `Key` on core groups/links (`DashboardLink`, `AdministrationLink`, children).
- Provide a registry-level **exclude/override** API (in `pkg/middleware/sidebar.go` / `pkg/sidebar/`) so a host can drop or replace a contributed group **by key** without inspecting its children.
- Guardrail test: core's nav contribution is deterministic and every contributed group carries a stable `Key`.

**Acceptance:**
- Hosts can suppress/override any SDK nav group by `Key` regardless of child shape.
- Changing a core group's children does not alter host sidebars that didn't opt in.
- `Key` is additive; existing `NavigationItem` literals without it still compile.

---

# Part C — Replace the tab-group builder with a declarative workspace model

## Problem (current mechanism)

Top-bar tabs (in the EAI host: ERP / CRM / EDO) are built by:

1. A **package-level mutable global** `sidebar.BuildTabGroups` (`pkg/sidebar/config.go:15`), reassigned by the host in `main()`. Not composition, not DI, one per process, awkward to test.
2. **Tab membership inferred by magic-href string match** (`pkg/sidebar/config.go:41` `item.Href == "/crm"`; the host override adds `"/edo"`). A nav group "becomes" a tab only because its href happens to equal a magic string. Worse, the meaning is inconsistent: `/crm` is a **fake marker** (no controller exists at `/crm`; the container is not a page), while `/edo` is a **real page**. One field, two meanings.
3. **Tab labels are not localized.** `TabGroupBuilder` receives a `*i18n.Localizer` but both the SDK default (`config.go:37` `_ *i18n.Localizer`) and the EAI override ignore it and hardcode `"Core"/"ERP"/"CRM"/"EDO"`. Latent i18n bug in a multi-locale app.
4. **Hosts must reimplement the whole builder to add a tab.** The SDK default produces 2 tabs (`Core`, `CRM`); the EAI host copy-extended it to 3 (`ERP`, `CRM`, `EDO`) and renamed Core→ERP. The default and the override have diverged.

### Usage finding that informs the design

Every workspace's nav is **already cleanly nested under a single container** (`CRMLink`'s 9 children, `EdoLink`'s children). The data is already grouped; only the *recognition* of a container as a tab is hacky.

### Types (`pkg/types/navigation.go`)

```go
type NavigationItem struct {
    Key         string                  // Part B
    Workspace   string                  // Part C: workspace/tab key; inherited by children; "" => Default workspace
    Pinned      bool                    // Part D
    Name        string
    Href        string
    Children    []NavigationItem
    Keywords    []string
    Icon        templ.Component
    Permissions []permission.Permission
    IsBeta      bool
}

// NavWorkspace declares a top-bar workspace (tab). Pure data, localizable.
type NavWorkspace struct {
    Key     string // stable id; matches NavigationItem.Workspace
    Label   string // i18n key, localized at render
    Order   int    // display order
    Default bool   // catch-all for items with empty/unknown Workspace
    IsBeta  bool
}
```

### Contribution API (`pkg/composition`)

```go
composition.AddNavWorkspaces(b,
    types.NavWorkspace{Key: "erp", Label: "NavWorkspaces.ERP", Default: true, Order: 0},
    types.NavWorkspace{Key: "crm", Label: "NavWorkspaces.CRM", Order: 1},
    types.NavWorkspace{Key: "edo", Label: "NavWorkspaces.EDO", Order: 2},
)
```

Items declare membership on the top-level container; children inherit:

```go
CRMLink.Workspace = "crm"   // no more Href=="/crm" marker
EdoLink.Workspace = "edo"
// everything else → Default workspace (erp). No per-item tagging.
```

### Builder

The SDK's default builder becomes **workspace-aware**: it reads the declared workspaces from the registry, partitions the (already permission-filtered, single-child-collapsed) nav items by inherited `Workspace`, sends unassigned items to the `Default` workspace, **localizes labels via the passed `*i18n.Localizer`**, and returns the `TabGroupCollection` (ordered by `Order`, empties dropped). **The host writes no builder.**

- Deprecate the mutable global `BuildTabGroups`. Keep it as an optional escape hatch for fully custom hosts (honored if set), but it is no longer required — the declarative registry is the primary path. (Non-breaking: hosts that still assign it keep working.)
- `getEnabledNavItems` single-child collapse and `filterItems` permission filtering run **before** workspace partitioning — unchanged.

### Why a workspace *field* (the resolved fork)

Chosen over (i) tagging every leaf item with a section, and (ii) a builder-scoped `composition.NewNavTab(b, key, label)` + `AddNavItems(tab, ...)` API. Rationale: real usage already nests each workspace's items under one container, so an inherited field on the container means migration = "tag 2 containers," not "re-annotate ~70 items" or "restructure module registration." Alternative (ii) is cleaner if we want modules to register directly into tabs at composition time — open to it if preferred; it's a larger module-registration change.

**Acceptance:**
- Adding a tab = one `NavWorkspace{}` entry; no builder edits.
- Tab labels localized across all locales.
- No magic-href coupling: renaming a route does not affect tab membership.
- Hosts with no declared workspaces get a single default list (today's single-tab behavior; tabs hidden when ≤1 group — `components/sidebar/sidebar.templ:80`).
- `Workspace` is additive; existing items compile unchanged.

---

# Part D — Always-visible (pinned) items above the workspace tabs

## Goal

A set of nav items that stay visible **regardless of which workspace (ERP/CRM/EDO) is active** — rendered above the per-tab content.

## Design

- `NavigationItem.Pinned bool` (or `composition.AddPinnedNavItems(b, items...)`) marks items as pinned; they are excluded from workspace partitioning.
- Carry them on the sidebar model: add `PinnedItems []Item` to `components/sidebar.Props` (or to `TabGroupCollection`).
- **Render slot:** in `components/sidebar/sidebar.templ`, insert a pinned `<nav>`/`<ul>` region between the header+tab-bar block (ends ~line 97) and `@SidebarNav(props)` (line 99) — i.e. **outside** every `navtabs.Content(tab.Value)`, so it does not switch with the active tab. Honors collapsed/flyout rendering like normal nodes.
- Permission filtering applies to pinned items as to any nav item.

**Acceptance:**
- Pinned items visible under every workspace tab and when only one workspace exists.
- Collapsed-sidebar + mobile drawer render pinned items correctly.
- Empty pinned set → no empty region rendered.

---

## Affected areas

- `pkg/types/navigation.go` — `Key`, `Workspace`, `Pinned`, `NavWorkspace`
- `modules/core/component.go` — `SkipAdminNavItems`; set `Key` on core nav
- `modules/core/links.go` — `Key` on core groups
- `pkg/composition/` — `AddNavWorkspaces`, `AddPinnedNavItems`; registry plumbing
- `pkg/sidebar/config.go` — workspace-aware default builder; deprecate global `BuildTabGroups`
- `pkg/middleware/sidebar.go` — partition by workspace + pinned; exclude/override-by-`Key`; localize labels
- `components/sidebar/types.go`, `view_model.go`, `sidebar.templ` — `Props.PinnedItems` + pinned render region
- Other consumers to keep working: `modules/core/presentation/controllers/showcase_controller.go`, `settings_controller_test.go` (construct `TabGroupCollection` directly — unaffected), BiChat `NavItemsWithInitialState`.

## Sequencing

- **A + B** are independent and ship first (unblocks EAI hotfix-debt removal).
- **C + D** build on B's `Key` and land together (same subsystem, same template).

## Non-breaking guarantees

- All new fields/structs are additive with behavior-preserving defaults.
- Hosts that set `BuildTabGroups` or build `TabGroupCollection` directly continue to work.

## Downstream

Host re-wiring (set `SkipAdminNavItems`, declare workspaces, tag containers, add pinned items, delete the custom builder + subtraction code) is tracked in **iota-uz/eai#2941**.

diyor28:
## Summary

- Adds stable navigation item keys plus workspace and pinned metadata.
- Adds composition APIs for declaring nav workspaces and removing/replacing nav items by key.
- Builds sidebar tab groups from declared workspaces with localized labels, while preserving the legacy `BuildTabGroups` hook when no workspaces are declared.
- Adds a pinned sidebar render region and `SkipAdminNavItems` for core admin nav decoupling.

Closes iota-uz/iota-sdk#791.
Unblocks iota-uz/eai#2941.

## Base branch

This PR targets `feat/archival-portfolio-import-2856` because EAI currently pins SDK commit `dd67fd82` from that branch. Targeting this branch keeps the downstream EAI build on the import APIs it already consumes; the navigation commit can be carried forward when that SDK branch lands on `main`.

## Validation

- `go test ./pkg/sidebar ./pkg/composition ./pkg/application ./modules/core ./components/sidebar`
- `go vet ./pkg/sidebar ./pkg/composition ./pkg/application ./modules/core ./components/sidebar`
- `git diff --check`

Note: a broader `go test ./pkg/middleware` run compiled the package but hit local Postgres because `TestPgxPoolCollector_CollectsFromRealPool` expects an `iota_erp` database on `127.0.0.1:5432`.

## Release Notes

* **New Features**
  * Added support for pinned navigation items in the sidebar for quick access to frequently used sections.
  * Implemented workspace-aware navigation organization, allowing navigation items to be grouped by workspace.
  * Enhanced collapsed sidebar mode with a compact workspace selector displaying workspace labels as abbreviated badges with tooltips.

* **Improvements**
  * Refined role management delete confirmation dialog to improve button interaction precision.
  * Added capability to independently control visibility of admin navigation items.

### DIFF
--- a/components/sidebar/sidebar.templ
+++ b/components/sidebar/sidebar.templ
@@ -19,6 +19,7 @@ const (
 type Props struct {
 	Header       templ.Component
 	TabGroups    TabGroupCollection
+	PinnedItems  []Item
 	Footer       templ.Component
 	InitialState SidebarState // Server-side initial state hint
 }
@@ -107,8 +108,22 @@ templ Sidebar(props Props) {
 }
 
 templ SidebarNav(props Props) {
-	{{ tabs := BuildSidebarNavTabs(ctx, props.TabGroups) }}
+	{{
+		tabs := BuildSidebarNavTabs(ctx, props.TabGroups)
+		pinnedNodes := buildSidebarNavNodes(ctx, props.PinnedItems)
+	}}
 	<nav x-data="sidebarNavigation()" x-init="initSidebarNavigation()" class="py-4 flex-1 min-h-0 overflow-y-auto hide-scrollbar">
+		if len(pinnedNodes) > 0 {
+			<ul
+				id="sidebar-pinned-navigation"
+				:class="{ 'px-2': isCollapsed, 'px-6': !isCollapsed }"
+				class="flex flex-col gap-2 pb-4 mb-4 border-b border-surface-300 transition-all duration-300"
+			>
+				for _, node := range pinnedNodes {
+					@SidebarNode(node, SidebarRenderModeMain, 0)
+				}
+			</ul>
+		}
 		for _, tab := range tabs {
 			@navtabs.Content(tab.Value) {
 				<ul
--- a/components/sidebar/sidebar.templ
+++ b/components/sidebar/sidebar.templ
@@ -150,11 +150,12 @@ templ SidebarPinnedNav(props Props) {
 // label) that switches the active tab via the same navTabs Alpine state used by
 // the expanded selector, so the two stay in sync.
 templ SidebarCollapsedWorkspaceTabs(props Props) {
-	<div x-show="isCollapsed" class="mt-4 flex flex-col items-center gap-1">
+	<div x-show="isCollapsed" role="tablist" aria-orientation="vertical" class="mt-4 flex flex-col items-center gap-1">
 		for _, group := range props.TabGroups.Groups {
 			<button
 				type="button"
 				role="tab"
+				aria-label={ group.Label }
 				data-tab-value={ group.Value }
 				@click={ fmt.Sprintf("setActiveTab('%s')", group.Value) }
 				x-bind:aria-selected={ fmt.Sprintf("isActive('%s') ? 'true' : 'false'", group.Value) }
--- a/components/sidebar/sidebar.templ
+++ b/components/sidebar/sidebar.templ
@@ -1,10 +1,12 @@
 package sidebar
 
 import (
+	"fmt"
 	icons "github.com/iota-uz/icons/phosphor"
 	"github.com/iota-uz/iota-sdk/components/base/button"
 	"github.com/iota-uz/iota-sdk/components/base/navtabs"
 	"strconv"
+	"strings"
 )
 
 // SidebarState represents the initial state of the sidebar
@@ -50,6 +52,19 @@ func depthAttr(depth int) string {
 	return strconv.Itoa(depth)
 }
 
+// workspaceTabBadge derives a compact, icon-sized label for the collapsed
+// workspace switcher. Short labels (e.g. "ERP", "CRM", "ЭДО") are shown as-is;
+// longer labels are truncated to their first 3 runes so they fit the 4rem rail.
+// The full label is always available via the button's tooltip.
+func workspaceTabBadge(label string) string {
+	label = strings.TrimSpace(label)
+	runes := []rune(label)
+	if len(runes) <= 3 {
+		return strings.ToUpper(label)
+	}
+	return strings.ToUpper(string(runes[:3]))
+}
+
 templ Sidebar(props Props) {
 	<div
 		x-data="sidebarShell()"
@@ -78,6 +93,7 @@ templ Sidebar(props Props) {
 				if props.Header != nil {
 					@props.Header
 				}
+				@SidebarPinnedNav(props)
 				if len(props.TabGroups.Groups) > 1 {
 					<div x-show="!isCollapsed" x-transition>
 						<div class="mt-4">
@@ -95,6 +111,7 @@ templ Sidebar(props Props) {
 							}
 						</div>
 					</div>
+					@SidebarCollapsedWorkspaceTabs(props)
 				}
 			</div>
 			@SidebarNav(props)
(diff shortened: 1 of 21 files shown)


diyor28:
Pushed follow-up work (`0bafd96c0`), driven by local review + browser QA of the downstream EAI PR (iota-uz/eai#2950):

**UI**
- Pinned nav items now render in the sidebar header **above** the workspace tab selector (were below), with a divider; stay visible when collapsed.
- New **collapsed-state workspace switcher**: when the sidebar is collapsed the horizontal tab bar is hidden, so each workspace renders as a compact icon-sized pill (badge label + right-placed tooltip) that switches the active tab via the same `navTabs` state — expanded and collapsed selectors stay in sync. Verified switching + two-way sync in the browser, no console errors.

**Review-finding fixes**
- `WorkspaceTabGroupBuilder`: empty-slice guard + deterministic secondary sort tie-break on `Key`.
- `localizeWorkspaceLabel`: now uses `intl.MustLocalize` (fail loudly on a missing `NavWorkspaces.*` key) to match nav-item-name localization semantics instead of silently shipping the raw key.
- `applyNavItemOverrides`: now recurses removals/overrides into replace-by-key replacement subtrees (previously ignored).
- middleware sidebar: stop aliasing the pinned-items slice backing array.
- **CI fix:** `pkg/itf/harness_internal_test.go` `testApp` mock was missing `NavWorkspaces()`, so `go test ./pkg/itf/...` did not compile — added it.
- Added tests: `splitPinnedItems`, `getEnabledNavItems`, `SkipAdminNavItems`, recursive overrides. Added a small read-only `composition.InspectStatic` helper to test static contributions without a DB (public API — flag if you'd prefer it unexported).

Also merged latest `main` into the branch (only #794 roles/edit fix; no conflicts). `go vet ./...` clean; targeted tests pass (pre-existing `TestPgxPoolCollector_CollectsFromRealPool` still needs a local Postgres).

diyor28:
Follow-up (`e474d5c91`): addressed the remaining auditor finding (#4) — `WorkspaceTabGroupBuilder` previously flattened **all** children of a workspace-tagged container into that container's workspace, discarding any child that declared a different `Workspace`. It now routes each child by its own (valid) `Workspace` when set, inheriting the container's workspace otherwise. Added `TestWorkspaceTabGroupBuilderHonorsChildWorkspaceOverride`. `go vet ./...` clean; `pkg/sidebar` tests pass.

diyor28:
@coderabbitai review

diyor28:
@coderabbitai addressed all 3 comments in `8671c1352`:

1. **a11y (collapsed workspace tabs):** wrapped the pills in `role="tablist"` + `aria-orientation="vertical"` and added `aria-label={ group.Label }` to each `role="tab"` button, so the full workspace label is exposed to assistive tech (not just the 3-char badge / tooltip).
2. **`localizeWorkspaceLabel` (critical):** reverted from `intl.MustLocalize` to a graceful `localizer.Localize` + raw-key fallback. Agreed — workspace labels are consumer-declared i18n keys, so a missing/typo'd `NavWorkspaces.*` key must not panic and take down the authenticated sidebar. Consumers validate keys via their own i18n checks.
3. **middleware test:** renamed the subtest to "single-child group chain collapses to grandchild leaf" and fixed the comment to match the actual fixture (parent → g1 group → g1a leaf).

`templ generate`, `go vet ./...`, and `pkg/sidebar` / `pkg/middleware` / `pkg/composition` / `modules/core` tests all pass.

PR #801, opened 2026-06-09
feat(scaffold/table): deferred summary panels (WithDeferredPanels)

diyor28:
## What

Adds a generic **deferred-panel** primitive to the scaffold table component so any table can render summary/aggregate/count regions that don't block first paint.

A `DeferredPanel`:
- paints a **skeleton** immediately (no blocking query),
- self-loads via `hx-get` on `load`,
- inherits the table's current filter/search querystring via `hx-include="closest form"`,
- **reloads** when the table form re-broadcasts its `RefreshEvent` after its *own* successful request,
- persists across row/content swaps (rendered as first child of the form, outside the swap target),
- uses `hx-sync="this:replace"` to cancel stale requests on rapid filter changes.

The `event.detail.elt===this` guard on the form's `hx-on::after-request` keeps the re-broadcast **loop-free** and naturally **excludes infinite-scroll and sort** from panel reloads (totals are order/pagination-invariant).

## Public API
- `DeferredPanel{ID, URL, Skeleton, Class}`
- `WithDeferredPanels(...DeferredPanel) TableConfigOpt`
- `(*TableConfig).RefreshEvent() string` — canonical `tbl:{ID}:refresh`
- `(*TableConfig).FormHxAttrs() templ.Attributes`
- templ: `DeferredPanels`, `DefaultPanelSkeleton`

## Why
Heavy list pages (e.g. EAI portfolio policies at ~728K rows) ran full-set `COUNT`/`SUM` aggregates **synchronously before first paint** and on every filter change. Deferring them to a skeleton-loaded panel removes ~1.3s+ from the critical path. First consumer: EAI `/portfolio/policies` (separate PR).

## Tests
- `deferred_panel_test.go`: `RefreshEvent` naming, `WithDeferredPanels`, `FormHxAttrs` (loop-free guard + event name), and a render test asserting the exact `hx-*` contract.
- Full `components/scaffold/table` suite green; additive change (no existing consumers affected).

## Notes
- `templ generate` run with pinned `templ@v0.3.857`.

## Release Notes

* **New Features**
  * Added deferred panels to tables—panels now persist outside the main table content and automatically reload when filters or search updates are applied. Loading skeletons display while content is being fetched.

* **Tests**
  * Added comprehensive unit tests for deferred panel behavior and table refresh events.

### DIFF
--- a/components/scaffold/table/deferred_panel_test.go
+++ b/components/scaffold/table/deferred_panel_test.go
@@ -1,0 +1,62 @@
+package table
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRefreshEvent(t *testing.T) {
+	assert.Equal(t, "tbl:orders:refresh", (&TableConfig{ID: "orders"}).RefreshEvent())
+	// Falls back to a stable name when ID is empty.
+	assert.Equal(t, "tbl:default:refresh", (&TableConfig{}).RefreshEvent())
+}
+
+func TestWithDeferredPanels(t *testing.T) {
+	cfg := NewTableConfig("Orders", "/orders",
+		WithID("orders"),
+		WithDeferredPanels(DeferredPanel{ID: "orders-summary", URL: "/orders/summary"}),
+	)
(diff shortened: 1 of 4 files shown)


diyor28:
First consumer: iota-uz/eai#3002 (defers the policies summary aggregates off first paint at 728K-row scale).

Issue #823, opened 2026-06-20
Tech debt: unresolved bot-review threads from merged PRs #798–#821


diyor28:
## Summary

Audit of the **20 most-recently merged PRs** (#798–#821) cross-referenced every code-review thread's timestamp and resolution state against each PR's merge time.

**Finding:** PRs are self-merged minutes after automated reviewers post (1.5–15 min windows), so bot feedback lands unresolved and becomes silent tech debt.

| Metric | Value |
|---|---|
| PRs merged with ≥1 unresolved thread | **13 / 20** |
| Unresolved review threads (pre-merge) | **44** |
| Thread authors | Copilot 24, CodeRabbit 20 (0 human) |
| Still pointing at live code | 39 / 44 |
| Verdicts | ✅ Fixed 16 · ☑️ Already-fixed 5 · ⬜ Backlog 20 · 🚫 Won't-fix 3 |

Legend: ✅ fixed by companion PR · ☑️ already addressed on `main` · ⬜ open backlog · 🚫 nitpick/won't-fix

### filterq / filterbuilder
- [x] ✅ **PR #806** `pkg/filterq/presets.go` — [PresetNext30D currently produces a 31-day inclusive range (from=today, to=today+30). This is inconsistent with PresetLast30D (30 days inclusive via -2](https://github.com/iota-uz/iota-sdk/pull/806#discussion_r3404302354) _( PresetNext30D off-by-one → 30-day span )_
- [ ] ⬜ **PR #806** `pkg/filterq/presets_test.go` — [This test encodes the current off-by-one behavior for PresetNext30D (to is 30 days after from, which yields 31 days inclusive). If PresetNext30D is in](https://github.com/iota-uz/iota-sdk/pull/806#discussion_r3404302394) _( needs follow-up triage )_
- [ ] ⬜ **PR #806** `components/scaffold/filterbuilder/builder_test.go` — [⚠️ Potential issue \| 🟠 Major](https://github.com/iota-uz/iota-sdk/pull/806#discussion_r3404364755) _( needs follow-up triage )_
- [ ] ⬜ **PR #806** `components/scaffold/filterbuilder/builder.templ` — [⚠️ Potential issue \| 🟡 Minor \| ⚡ Quick win Hide empty field groups when search filters out every item. The search predicate is only applied to fieldPi](https://github.com/iota-uz/iota-sdk/pull/806#discussion_r3404364779) _( needs follow-up triage )_
[...]

PR #819, opened 2026-06-20
feat(copy_button): add CopyableText click-to-copy component

diyor28:
## Summary
Adds `CopyableText` to the `copy_button` package — the canonical click-to-copy element for short identifiers (policy/contract numbers, INN, PINFL, passport, external IDs) in tables.

- **Hover-reveal** copy icon next to the value; click copies to clipboard and swaps the icon for a green check (~1.5s).
- **Whole element is the click target** and uses `@click.stop`, so copying does not also fire a row-level handler (e.g. a drawer-opening table row).
- **Injection-safe**: the value is carried in a `data-copy-text` attribute and read via `$el.dataset` at click time, never interpolated into the Alpine handler. The same hardening is applied to the existing `CopyButton`.
- Empty values render as a `-` placeholder; colors inherit the surrounding text color.

## API
```go
copy_button.CopyableText(copy_button.TextProps{
    Text:  value,          // displayed and copied
    Label: localizedHint,   // tooltip / aria-label (SDK has no i18n; caller passes a translated string)
})
```

## Consumer
EAI adopts this for new copy affordances on claims / persons / organizations tables and migrates its in-house `pkg/templates.CopyableText` onto it (linked PR).

## Test plan
- `templ generate && go vet ./components/copy_button/...` clean.
- Verified in EAI against a prod-scale dataset (Chrome): hover reveals the icon, click copies + shows the green check, and on drawer-row tables (`/crm/lookups/persons`) clicking copies **without** opening the drawer.

https://claude.ai/code/session_016psCQKbeZbijyrsDyXJKhu

## Release Notes

* **New Features**
  * Added a new CopyableText component for inline, clickable copy functionality with improved visual feedback.

* **Improvements**
  * Enhanced accessibility with better labeling and ARIA attributes for copy interactions.
  * Updated copy button behavior with more reliable data handling and faster visual feedback reset.

PR #824, opened 2026-06-20
fix: resolve unaddressed bot-review feedback (tech-debt batch)

diyor28:
## Context

Audit of the 20 most-recently merged PRs (#798–#821) found **13 merged with unresolved automated-review threads** (Copilot/CodeRabbit), typically self-merged 1.5–15 min after the bot posted. Full triage of all 44 threads is tracked in #823.

This PR fixes the genuine bugs surfaced by that feedback. Each has a regression test.

## Changes

| Area | Fix | Source thread |
|---|---|---|
| `pkg/filterq/presets.go` | `PresetNext30D` returned a 31-day inclusive range → corrected to 30-day span (symmetric with `PresetLast30D`) | #806 |
| `pkg/filterq/codec.go` | `validValue` accepted `NaN`/`Inf` for number fields → now rejected | #806 |
| `pkg/filterq/codec.go` | Multi-valued boolean conditions (`is:true,false`) → now rejected | #806 |
| `components/scaffold/filterbuilder/registry.go` | `NewRegistry` kept all fields but a last-wins index, so `Fields()`/`Field()`/`Schema()` could disagree on duplicate keys → dedupe first-wins | #806 |
| `pkg/middleware/htmx_cache.go` | `HTMXCacheControl` skipped cache-safety headers when the handler left `Content-Type` unset → now sniffs body via `http.DetectContentType` like net/http | #809 |
| `components/copy_button/copy_button.templ` | `CopyableText` not keyboard-accessible and flipped `copied` without awaiting clipboard → added `role`/`tabindex`/keydown + awaited-and-caught handler | #819 |
| `modules/core/presentation/assets/js/alpine.js` | combobox `destroy()` leaked the `<select>` `MutationObserver` → now disconnected | #816 |

## Out of scope
- Root-cause process gate (branch protection requiring thread resolution) — deferred, noted in #823.
- ~20 remaining backlog threads + nitpicks — tracked in #823.

## Verification
- `go vet ./...` — clean
- `go test ./pkg/filterq/... ./components/scaffold/filterbuilder/...` — pass
- `go test ./pkg/middleware/ -run TestHTMXCacheControl` — pass (5 new cases incl. empty-Content-Type sniff)
- `templ generate && just css` — clean

Tracks #823

https://claude.ai/code/session_01QbyJYndqNDnPfPCjeaUFoF

## Release Notes

* **New Features**
  * Copy button now supports keyboard activation via Enter and Space keys with enhanced focus indicators for better accessibility.

* **Bug Fixes**
  * Fixed "next 30 days" date filter to span exactly 30 days instead of 31.
  * Improved filter validation to reject invalid numeric values and malformed condition inputs.

* **Improvements**
  * Enhanced clipboard error handling.
  * Refined caching behavior for better performance on HTML responses.


diyor28:
Addressed the review feedback in 38faf89:

- **htmx_cache `WriteHeader`-first + empty Content-Type (Copilot & CodeRabbit, Major):** Confirmed real against a *real* `httptest.NewServer` — net/http sniffed `text/html` but `Vary`/`Cache-Control` were dropped post-commit. The original test used `ResponseRecorder`, which never freezes its header map, so it masked the bug. Fixed by buffering the status until Content-Type is known, then committing once. Added a real-server regression test + a buffered-204 case. (Used scoped buffering rather than the suggested unconditional `no-store` default, which would over-apply `no-store`/`Vary` to non-HTML responses served without an explicit Content-Type, e.g. JSON-via-sniff.)
- **filterq bool merge bypass (Copilot, Major):** Confirmed — the single-value guard ran before `Decode`'s duplicate-merge, so `is:true` + `is:false` merged into a match-everything condition. Now bool fields don't merge (first wins). Regression test added.
- **CopyableText timeout 1500 vs 2000ms (CodeRabbit, nitpick):** Skipped — 1500ms is CopyableText's existing UX; not worth the churn.

All targeted tests + `go vet ./...` pass.

Issue #739, opened 2026-04-13
Big-bang rewrite controller override model: separate controller identity from routing


diyor28:
## Problem

`Controller.Key()` is currently overloaded. It is used as:
- the composition removal target
- the duplicate controller detection key
- the controller sort key during registration
- the de facto route identifier in most controllers

This creates a recurring class of bugs:
- `RemoveController(builder, "/")` can remove both upstream and downstream controllers when both use `/`
- downstream overrides are forced to invent fake keys like `/-eai-dashboard` or `/settings-eai` just to survive composition
- once fake keys are introduced, the framework no longer has a truthful view of the actual mounted routes and cannot reliably detect surviving route collisions

Examples already observed in EAI include dashboard, settings, and logout override failures.

## Goal

Replace the current key-based controller composition with an explicit controller identity and override model.

This should be a full big-bang rewrite. No deprecation shims, adapters, compatibility layers, or legacy `Key()` fallback.

## Proposed direction

Replace `Controller.Key()` with a richer descriptor, for example:

```go
type Controller interface {
    Register(r *mux.Router)
    Descriptor() ControllerDescriptor
}

type ControllerDescriptor struct {
    ID       string
    Order    int
    Replaces []string
    Routes   []RouteSpec
}
```

Where:
- `ID` is the stable controller identity and is never derived from mux path
- `Order` controls registration order instead of sorting by string key
- `Replaces` explicitly declares which controller IDs are being overridden
- `Routes` (or equivalent metadata) is used for startup-time validation of duplicate surviving route patterns

Composition behavior should become:
- no `RemoveController(builder, "...path...")`
- fail if two surviving controllers share the same `ID`
- fail if two surviving controllers declare conflicting routes unless explicitly allowed by the framework design

## Scope

- rewrite the controller interface and controller composition API in SDK
- migrate all SDK controllers, including applet controllers
- update sorting/registration logic so it no longer relies on `Key()` or magic prefixes like `applet_`
- replace current controller-removal tests with explicit override/replace tests based on controller IDs
- coordinate downstream migrations (EAI and other consumers) in the same rollout window

## Acceptance criteria

- there is no `Controller.Key()` API in SDK
- there is no path-string-based `RemoveController(...)` API
- controller replacement is done through explicit controller IDs
- controller registration order is configured explicitly, not by lexicographic key sorting
- duplicate surviving route collisions are detected by the framework
- downstream apps can override SDK controllers without invented composition keys
[...]

PR #826, opened 2026-06-21
[codex] rewrite controller descriptors

diyor28:
## Summary
- replace controller Key() with Descriptor() across the SDK
- add route metadata helpers and descriptor-backed controller ordering/replacement
- validate duplicate IDs and cross-controller route collisions at composition startup
- migrate superadmin dashboard override to Descriptor().Replaces and update docs/tests
- bump github.com/iota-uz/applets to the descriptor-capable pseudo-version

## Dependency
- Depends on https://github.com/iota-uz/applets/pull/32 being merged or tagged before this SDK PR is merged/released.

## Validation
- go test ./... in github.com/iota-uz/applets
- go test -v ./pkg/composition ./pkg/application ./pkg/appletengine/controllers ./modules/bichat/presentation/controllers
- go test -run ^\\ ./...\n- go vet ./...\n- rg -n "RemoveController|Controller\\.Key|controller key|controller-removal|controllerRemovals" . -g '!*_templ.go' (no matches)\n\nCloses #739


diyor28:
## RFC: extend descriptors so navigation, spotlight & route-authz are *derived*, not hand-mirrored

Building directly on this PR's core move — making `Descriptor()` the source of truth for **routes** — I want to fold in a second architectural fix that this PR is the natural home for. This was stress-tested with an adversarial review (Codex, read-only against the tree); the design below is the **post-review** version and several first-draft decisions were reversed by it.

### Problem

A single navigable feature is declared in **3+ unsynchronized places** with no compiler link between them:

| Place | Declares | Drift / failure mode |
|---|---|---|
| `Descriptor().Routes` (this PR) | the route | source of truth ✅ |
| `modules/*/links.go` `NavigationItem{Href, Permissions, …}` | sidebar entry — `Href` is a hand-copied route string, `Permissions` hand-copied | silently diverges from the route |
| `modules/*/component.go` `AddQuickLinks(…)` | spotlight — `Name`+`Href` copied **again** | forget it → feature unsearchable |

~55 nav items + ~39 quick-links across 8 modules, all hand-maintained mirrors. Add a screen → touch 3–4 files; miss one → broken sidebar / missing spotlight / wrong perms. **Nothing validates that a nav `Href` resolves to a real route.**

And a deeper issue the review surfaced: **routes have no declarative authorization at all.** `middleware.Authorize()` only *authenticates*; permission checks are imperative inside handlers (`debt_controller.go:96`), and the sidebar's `NavigationItem.HasPermission` is **display-only** — a direct URL hit bypasses it. So today permissions are *also* duplicated and *also* unenforced at the route.

### Principle

> **One declaration per surface. Routes, nav, spotlight, sitemap, command-palette and route-authz all project from the controller descriptor + a module-level group catalog, cross-validated at composition (fail-fast) — never authored in parallel.**

### Shape

```
RouteSpec.Auth ─────────────► middleware enforces (closes URL-bypass gap)
        │
ControllerDescriptor.Nav ──┐
   (leaf nodes, w/ Path)    ├──► engine.buildNavModel()  ── validates ──►  projects
NavCatalog (group nodes) ──┘     • every Path resolves to a route          ┌─ Sidebar (NavigationItem tree)
NavProvider (runtime) ─────►     • visibility ← route Auth (not nav)       ├─ Spotlight quick-links
   (tenant/role/DB-driven)       • dup ID / orphan / cycle → panic         ├─ Sitemap
                                 • runtime nodes validated on emit →       └─ Command palette / mobile (future)
                                   dropped+logged, never panic
```

`types.NavigationItem` stays the **render** type (no template churn) but becomes *derived*. `spotlight.QuickLink` stays the index type, also derived.

### Types (public surface)

```go
// ── ControllerDescriptor: nav is a REQUIRED field; nil = intentionally non-navigable ──
type ControllerDescriptor struct {
    ID     string
    Order  int
    Routes []RouteSpec
    Nav    []NavNode      // explicit nil for webhooks / APIs / health — forces a conscious choice
}

// ── Route carries its own authz; middleware enforces it; nav derives visibility from it ──
type RouteSpec struct {
    Method, Path   string
    Prefix         bool
    Host           string
    AllowCollision bool
    Auth           AuthPolicy   // NEW
}
type AuthPolicy struct {
    Public      bool                    // login / webhooks / health
    Permissions []permission.Permission
    Logic       PermissionLogic         // Any | All  (sidebar today = All, spotlight = Any — unified here)
}

// ── ONE node type (leaf and group unified — fixes the "mapper drops Href when it has children" leak) ──
type NavNode struct {
    ID, Parent, Workspace string         // Parent "" => workspace root
    TitleKey  string                      // i18n key
    Path      string                      // optional href; query allowed (3 finance tabs → 1 route); matched path-without-query
    Icon      templ.Component
    Order     int
    Before, After string                  // ordering anchors across independently-developed modules
    Keywords  []string                    // i18n keys (uz/ru/oz), not raw strings
    Surfaces  map[Surface]SurfaceOptions  // empty => Sidebar+Spotlight; per-surface label/icon/order/hidden overrides
    Actions   []NavAction                 // quick-create ("New expense"); Spotlight by default
    IsBeta    bool
    Visibility *AuthPolicy                // groups w/o Path only; leaves inherit their route's Auth
}

// ── Runtime extension point for EDO / multi-tenant / DB-driven menus ──
type NavProvider interface {
    ProvideNav(ctx context.Context, scope NavScope) ([]NavNode, error)
}
```

**Ownership (review point K):** controllers own only their **leaf** nodes (route-adjacent). Workspace/group **taxonomy** stays module-level via a `NavCatalog` (`composition.AddNavNodes`) so controllers don't become cross-module IA coordinators.

### What module code becomes

```go
// controller — Path reuses the same basePath const as the route ⇒ cannot drift; visibility derived from Route.Auth
func (c *ExpenseCategoriesController) Descriptor() application.ControllerDescriptor {
    return application.Descriptor("finance.expense_category", 0,
        application.Route("GET", c.basePath, application.RequireAll("finance.expense_categories.view")),
    ).WithNav(application.NavNode{
        ID: "finance.expense_categories", Parent: "finance.enums",
        TitleKey: "NavigationLinks.ExpenseCategories", Path: c.basePath,
        Actions: []application.NavAction{{TitleKey: "ExpenseCategories.List.New", Path: c.basePath + "/new"}},
    })
}

// component.go — 3 blocks collapse to 1; AddNavItems + AddQuickLinks DELETED (both derived)
func (c *component) Build(b *composition.Builder) error {
    composition.AddNavNodes(b, finance.Groups...)   // grouping/taxonomy nodes only
    …controllers…
}
```

### Decisions (post-review)

| Decision | Choice | Why |
|---|---|---|
| Route authz | **`AuthPolicy` on `RouteSpec`, enforced in middleware; nav visibility derives from it** | the only correct "single source" — closes the direct-URL bypass; `NavNode.Permissions` as the source would have *codified* the bypass |
| Nav on controller | **required descriptor field, explicit `nil`** (not an optional interface) | an optional `Navigable` interface re-creates the exact silent-forget bug we're killing; a lint flags GET-HTML routes with no nav |
| Group vs leaf | **one `NavNode`** with optional `Path`/`Parent` | the split leaks — the sidebar mapper already drops `Href` for parents (`map.go:8`) while finance ships clickable parents (`/finance`) |
| Permission logic | **explicit `Any`/`All`** | sidebar requires All (`navigation.go:23`), spotlight allows Any (`acl.go:178`) — silent mismatch today |
| Dynamic nav | **static catalog + request-scoped `NavProvider`** | EDO (role/data-driven menus) + multi-tenancy can't be expressed statically; a decorator pipeline (`SetNavItemsDecorator`) already exists to build on |
| Surfaces | **`map[Surface]SurfaceOptions`**, not a bitmask | mobile/breadcrumbs/command-palette need per-surface label/icon/order |
| Validation | **fail-fast panic** for static (dup ID, unresolved Path, same-module orphan, cycle); **cross-module orphan** (optional dep absent) → skip+warn; **runtime** nodes → drop+log | reuses this PR's collision-validation pass; doesn't break optional-module setups |

### Rollout

1. **Engine + types** — `AuthPolicy`/`NavNode`/`NavProvider`/`Surface` in `pkg/application`; `buildNavModel` + validation + projections in `pkg/composition`; enforcement middleware reads `RouteSpec.Auth`.
2. **Migrate 8 modules** — leaves → `Descriptor().Nav`; `links.go` → `NavCatalog` group tables; delete `AddQuickLinks` blocks; move imperative handler perm-checks → `RouteSpec.Auth`.
3. **Delete dead paths** — `AddNavItems`/`AddQuickLinks` helpers + per-module nav var soup.
4. **Parity gate (merge-blocking)** — golden snapshots of the materialized `NavigationItem` tree, the mapped sidebar tree, **and** the quick-link doc set, captured before migration; diff must be empty (modulo intended fixes). The render pipeline has nontrivial behavior — pinned hoist, single-child collapse, workspace inheritance — that must not silently shift.

PR #835, opened 2026-06-24
[codex] Fix stacked chart tooltips and rerender state

diyor28:
## Summary
- add Lens measure-level dataset overrides for stat panels, mirroring existing dimension override behavior
- add a localized total row to Lens stacked-bar Apex tooltips
- preserve hidden legend series when charts are destroyed and re-rendered
- add Lens compile/cube tests plus Apex renderer coverage and regenerate chart templ output

## Validation
- `go generate ./... && templ generate`
- `go test ./components/charts ./pkg/lens/render/apex`
- `go test ./pkg/lens/...`
- `go test ./...` attempted earlier; failed in `pkg/middleware` because local PostgreSQL test database `iota_erp` is unavailable/refused, while prior packages passed

Parent: iota-uz/eai#3257

### DIFF
--- a/components/charts/chars.templ
+++ b/components/charts/chars.templ
@@ -10,20 +10,55 @@ import (
 )
 
 script graph(id string, options templ.JSExpression) {
+	const hiddenSeriesNames = (chart) => {
+		const globals = chart && chart.w && chart.w.globals;
+		if (!globals) {
+			return [];
+		}
+		const names = globals.seriesNames || [];
+		const hidden = new Set();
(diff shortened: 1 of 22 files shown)


diyor28:
Superseded by #839, which includes the chart-tooltip fix and measure-overrides (cleanly rebased onto current `main`) alongside the SegmentBar widget + drill-indicator scope fix. The incomplete restore-legacy `Descriptor` commits here are unnecessary — `main` already carries that migration.

PR #836, opened 2026-06-24
feat(excel): add streaming ExportToWriter for large exports

diyor28:
## What

Add `ExcelExporter.ExportToWriter(ctx, io.Writer, DataSource)` — a streaming Excel writer built on excelize's `StreamWriter`.

## Why

The existing `Export() ([]byte, error)` materializes the whole workbook in memory and then re-walks every row to apply styles (O(N) memory + an O(N) style pass). For large exports (hundreds of thousands of rows) this spikes memory and delays the first response byte until the entire file is built — which, behind a proxy with a response-header timeout, surfaces as a Gateway Timeout in the consuming app (EAI portfolio policies export).

## How

- `StreamWriter` spools rows to a temp file → peak memory stays flat regardless of row count.
- Number formats (money `0.00`, dates, ints) are attached per cell at write time, so money stays a real number, not text.
- Header style, column widths and numeric/date formats preserved.

## Trade-off

No alternate-row striping or per-cell borders (StreamWriter cannot style a range after a row is flushed). Callers pick by size: keep `Export` for small exports that want rich styling; use `ExportToWriter` for large ones. `Export` is unchanged.

## Tests

`TestExcelExporter_ExportToWriter` (headers + money-stays-numeric guard) and `TestExcelExporter_ExportToWriter_MaxRows`. `go test ./pkg/excel/` passes.

Consumed by EAI portfolio policies export remediation (paired eai PR).

## Summary by CodeRabbit

* **New Features**
  * Added a Python kernel manager with sandboxed execution, capability dispatch, lifecycle policies, and warm/ephemeral pooling.
  * Introduced streaming Excel exports and fully clickable table cells.
  * Added a new SQL capability for querying data from the Python kernel.
  * Added bridge support for host/kernel communication and Unix socket-based kernel connections.

* **Bug Fixes**
  * Improved process cleanup, environment isolation, and resource-limit handling for managed Python kernels.

* **Documentation**
  * Added and updated documentation for the Python kernel bridge and deployment image.

### DIFF
--- a/pkg/excel/exporter.go
+++ b/pkg/excel/exporter.go
@@ -4,6 +4,7 @@ package excel
 import (
 	"context"
 	"fmt"
+	"io"
 	"time"
 
 	"github.com/xuri/excelize/v2"
@@ -12,6 +13,8 @@ import (
 // Exporter exports data to Excel format
 type Exporter interface {
 	Export(ctx context.Context, datasource DataSource) ([]byte, error)
+	// ExportToWriter streams the export to w with flat memory (large exports).
+	ExportToWriter(ctx context.Context, w io.Writer, datasource DataSource) error
 }
 
 // ExcelExporter implements Exporter using excelize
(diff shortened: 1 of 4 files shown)


diyor28:
Superseded by #837. The excel change is independent of the pykernel/scaffold stack this branch was based on, so it now targets main directly as a clean, lint-green PR. This branch (claude/jovial-mendel-0fa6be) stays as the interim SDK pin for eai#3267 (it carries pykernel + scaffold + excel so eai builds today); eai re-pins to main once #837 + the pykernel/scaffold PRs land.

PR #839, opened 2026-06-24
feat(lens): SegmentBar + drill indicator scope fix + chart tooltip + measure overrides

diyor28:
## Summary

`pkg/lens` + chart/filter fixes over `main`, consumed by the EAI analytics, sales-report and portfolio dashboards. **Supersedes #835** (its chart-tooltip + measure-override commits are included here, cleanly rebased onto current `main`; #835's incomplete `Descriptor` restore-legacy detour is dropped — `main` already has the migration). Branch is merged up to current SDK `main`.

## Bugs fixed / changes

Each entry: **what the user saw → the actual root cause → the fix.**

### 1. Stacked-chart tooltip + rerender state _(from #835)_
Stacked-bar tooltip totals and chart rerender behavior were wrong. Corrected.

### 2. Lens measure dataset overrides _(from #835)_
`.Override(...)` at the **measure** level so a KPI stat card can use an authoritative precomputed frame while charts keep rendering from the bounded aggregate. Backs the EAI sales-report exact total-policies KPI.

### 3. `SegmentBar` part-to-whole widget _(new primitive)_
Headline total + one proportional bar with per-segment share + inline legend; clickable like `StatPanel`. `panel.KindSegmentBar` + `spec.SegmentBar(...)` + templ renderer + `runtime.Validate` support.

### 4. Drill `htmx-request` indicator leaked onto every button
**Saw:** opening a metric drawer flashed the loading state (transparent label + dots) onto every button on the page — nav tabs, sidebar. **Cause:** drill click actions called `htmx.ajax(verb, url, {target, swap})` with **no `source`**; HTMX falls back to `document.body` as the request element (`he()`: `if(r==null){r=ne().body}`), so `body.htmx-request` cascades the SDK button loading style onto every `.btn`. **Fix:** pass an explicit `source` (`source: this` for templ panels, `source: cfg.target` for the apex fallback), matching the cube-drill pattern.

### 5. Lens tabs — inactive tab text unreadable (low contrast)
**Saw:** the inactive tab label (e.g. «Количество») was near-black on the dark `bg-slate-800` tab nav — effectively invisible. **Cause:** inactive text used `text-slate-300`, but the consuming Tailwind build never generates `.text-slate-300` (only the `.dark:` variant), so the class matched no rule and the button fell back to `button { color: inherit }` → the page's near-black body color. **Fix:** use `text-slate-400` (a generated, contrast-verified muted-light shade) for the inactive tab.

### 6. Chart legend selection resets when switching tabs
**Saw:** deselecting a series in a `KindTabs` chart panel (e.g. hiding ОСАГО on «Доход») reset when switching to the sibling tab («Количество»). **Cause:** each tab renders a **separate ApexCharts instance** with its own hidden-series state — not a destroy/recreate loss; the sibling chart simply never knew about the deselection. **Fix:** share one hidden-series set across charts in the same `[data-lens-rerender-scope]` — written on legend click (a `legendClick` hook publishes the post-toggle set to the scope) and read at render time — so the selection persists and propagates across tabs in both directions (deselect and re-enable).

### 7. Date "between" filter silently dropped (no chip, not applied)
**Saw:** selecting a date range with the `between` operator added no filter chip and didn't filter results. **Cause:** the filter panel's `collect()` for `date` grabbed **every** hidden input in the variant, including flatpickr's altInput original (`.flatpickr-input`), which holds the joined `"from — to"` display string. A range therefore produced **3** values (display string + the two dates), failing the `between` arity-of-2 check, so the condition was discarded. **Fix:** exclude `.flatpickr-input` from the selector → exactly 2 values for a range (1 for single).

### 8. Combobox dropdown: collapsed to 100px + clipped long labels
**Saw:** the searchable combobox (e.g. analytics product picker) opened a tiny ~2.5-row dropdown, and long option labels were cut off horizontally. **Cause (height):** the base stylesheet carried a stray `:popover-open { width:200px; height:100px; position:absolute; inset:unset; bottom:5px; right:5px; margin:0 }` rule — the literal MDN popover-API demo snippet, introduced with the canonical Tailwind CSS input. Native `popover` is used only by the combobox dropdown, so this forced **every** open dropdown to a fixed 100px, overriding `positionDropdown`'s content-driven `max-height` (which is why an earlier height fix appeared not to land). **Cause (width):** `positionDropdown` pinned the menu to the trigger width, and options are `whitespace-nowrap` under `overflow-x:hidden`, so anything wider than the field clipped.

**Fix (this PR — width):** `positionDropdown` now grows the menu to the widest option, capped to the viewport, with a left-clamp so a wider menu never overflows the right edge. Short-option comboboxes are unchanged (still match field width). This is the JS half and lands in any consumer via the Go module.

**Fix (height):** the stray rule is removed from this repo's `styles/tailwind/iota.css` (gitignored compiled CSS → rebuilt in CI) for the SDK's own builds. Note: consumers like EAI bundle their CSS from the **npm** `@iota-uz/sdk` package (built from the applets repo), not this Go module — so the height fix reaches them via **iota-uz/applets#33** + the next `@iota-uz/sdk` release, not via this PR's pin.

### 9. Metric Info tooltip mounted in the card, not `document.body`
**Saw:** the ⓘ "how this metric is calculated" popover on lens Stat/chart panels bled through onto neighbouring cards — their numbers and borders showed over the popover, with stray line artifacts (reported on the EAI profitability dashboard). **Cause:** `MetricInfoButtonWithText` renders an *interactive* Tippy. Tippy's `appendTo` resolver has a special case — `(interactive && appendTo === DEFAULT) || appendTo === "parent"` → mount in the reference's `parentNode`. With the default `appendTo`, an interactive tooltip therefore mounts inside the card header, trapping its `z-index:9999` in the card's local stacking context; the overflow into a sibling card is then painted over by that later sibling. **Fix:** set an explicit `appendTo: () => document.body` on the instance (a distinct function reference, so it bypasses the interactive-parent special case) via `x-init` → `setProps`. Interactivity preserved; popover now mounts at body level. Verified in-browser (`[data-tippy-root].parentElement === document.body`, opaque over neighbouring content).

## Test plan
- `go build ./...` ✓
- `go test ./pkg/lens/... ./components/charts/...` ✓
- `templ generate` — no drift ✓
- Bugs 5–7 verified in-browser (Chrome) on the EAI sales-report + portfolio/policies dashboards: inactive tab readable; legend deselection carries across tabs both directions; `between` filter produces a valid 2-value condition (`policy_start_at:between:<from>,<to>`), renders a chip, and filters the result set.
- Bug 8 root-caused in-browser (dropdown measured at 256×100 with `max-height:1050` overridden by the stray `:popover-open` rule, options needing ~680px). **Width fix verified** — the menu now grows to fit full option labels (no clip). **Height fix** lands in EAI via iota-uz/applets#33 + the next `@iota-uz/sdk` npm release (EAI's CSS is bundled from the npm package, not this Go module).

## Downstream
Required by **iota-uz/eai#3270**, which pins this commit.

## Review feedback addressed (latest commits)

**CI Code Quality** was failing on two things, both fixed: (a) a gofmt misalignment in `pkg/lens/panel/panel.go` after `KindSegmentBar` was added, and (b) the `exhaustive` linter — `KindSegmentBar` was added to the `panel.Kind` enum without being handled in the many `switch` statements over it. Every such switch now handles `KindSegmentBar` in its behaviourally-neutral bucket (a SegmentBar renders via the `SegmentBarPanel` templ component, never the Apex/chart path, so behaviour is unchanged). `golangci-lint run --build-tags dev ./...` now reports 0 issues and all checks are green.

Automated review comments (CodeRabbit + Copilot), each verified against current code:

**Fixed**
- **TabsPanelFrames didn't route `KindSegmentBar`** → a SegmentBar nested in a Tabs panel fell through to the Apex chart path. Added the `KindSegmentBar → SegmentBarPanel` branch.
- **Category-only SegmentBar rendered `<nil>` labels** → validation accepts label *or* category but `buildSegmentBarView` only read `Fields.Label`; it now falls back to `Fields.Category` before the default.
- **Segment slice width unclamped above 100%** → `segmentSliceStyle` now clamps to `[0,100]`.
- **Chart legend-sync hazards** (`components/charts`): publish reads the event's `chartCtx` instead of the possibly-replaced `container.__apexChart`; the outgoing chart's hidden set is no longer merged over an authoritative shared tab-group selection; and the async `render()` callback bails if a newer render superseded it.

**Reviewed, intentionally not changed (with reason)**
- **`apex.go` `source: cfg.target`** — not a bug: `htmx.ajax` resolves a CSS-selector *string* for `source`, so the loading-state scoping works.
- **`apex.go` locale prefix casing** — `strings.ToLower` would *break* `uz-Cyrl` matching; `language.Tag.String()` already emits canonical-cased tags.
- **SegmentBar percentage denominator vs zero rows** — zero rows add 0 to the total and are hidden by the `Raw > 0` guard, so non-negative inputs still sum to 100%.
- **Error-context wrapping suggestions** — diagnostic-only; the underlying errors are already self-describing.

Verified after the changes: `go vet ./pkg/lens/...` + full `pkg/lens` tests pass, and in-browser against the EAI dashboards — SegmentBar renders (78.5% / 21.5% slices) and legend deselection still carries across tabs in both directions.

## Summary by CodeRabbit

* **New Features**
  * Added support for a new segment bar panel type with totals, legends, segmented progress, and clickable panel actions.
  * Added measure-level dataset overrides to power measure-specific data sources.
  * Kept chart legend series visibility synchronized across related charts during rerenders.

* **Bug Fixes**
  * Improved stacked bar tooltips to include per-series breakdown and a localized total (handling collapsed series).
  * Scoped HTMX swap loading/styling to the swap target element.
  * Improved dropdown sizing/positioning to stay within the viewport and fixed date range hidden input collection.

### DIFF
--- a/components/charts/chars.templ
+++ b/components/charts/chars.templ
@@ -10,20 +10,55 @@ import (
 )
 
 script graph(id string, options templ.JSExpression) {
+	const hiddenSeriesNames = (chart) => {
+		const globals = chart && chart.w && chart.w.globals;
+		if (!globals) {
+			return [];
+		}
+		const names = globals.seriesNames || [];
+		const hidden = new Set();
+		(globals.collapsedSeriesIndices || []).forEach((index) => {
+			if (names[index]) {
+				hidden.add(names[index]);
+			}
+		});
+		(globals.collapsedSeries || []).forEach((series) => {
+			if (typeof series === 'string') {
+				hidden.add(series);
+				return;
+			}
+			if (series && typeof series.name === 'string') {
+				hidden.add(series.name);
+			}
+		});
+		return Array.from(hidden);
+	}
+
 	const renderChart = () => {
 		const container = document.getElementById(id);
 		if (!container) {
 			console.error(`Chart container with ID ${id} not found.`);
 			return;
 		}
+		const hidden = new Set(container.__apexHiddenSeries || []);
 		if (container.__apexChart && typeof container.__apexChart.destroy === 'function') {
+			hiddenSeriesNames(container.__apexChart).forEach((name) => hidden.add(name));
 			container.__apexChart.destroy();
 			container.__apexChart = null;
 		}
 		container.innerHTML = '';
 		const chart = new ApexCharts(container, options);
 		container.__apexChart = chart;
-		chart.render();
+		container.__apexHiddenSeries = Array.from(hidden);
+		Promise.resolve(chart.render()).then(() => {
+			const names = ((chart.w && chart.w.globals && chart.w.globals.seriesNames) || []);
+			hidden.forEach((name) => {
+				if (names.includes(name) && typeof chart.hideSeries === 'function') {
+					chart.hideSeries(name);
+				}
+			});
+			container.__apexHiddenSeries = hiddenSeriesNames(chart);
+		});
 	}
 	document.addEventListener('DOMContentLoaded', () => {
 		renderChart();
--- a/components/charts/chars.templ
+++ b/components/charts/chars.templ
@@ -67,7 +67,10 @@ script graph(id string, options templ.JSExpression) {
 			// the resulting hidden set next tick and publish it to the shared scope
 			// for sibling charts to adopt on their next render.
 			setTimeout(() => {
-				const names = hiddenSeriesNames(container.__apexChart);
+				// Read the chart that fired this click (chartCtx), not
+				// container.__apexChart — a re-render may have replaced the latter
+				// by the time this tick runs, which would publish the wrong set.
+				const names = hiddenSeriesNames(chartCtx || container.__apexChart);
 				container.__apexHiddenSeries = names;
 				writeSharedHidden(container, names);
 			}, 0);
@@ -87,7 +90,13 @@ script graph(id string, options templ.JSExpression) {
 		const sharedHidden = readSharedHidden(container);
 		const hidden = new Set(sharedHidden !== null ? sharedHidden : (container.__apexHiddenSeries || []));
 		if (container.__apexChart && typeof container.__apexChart.destroy === 'function') {
-			hiddenSeriesNames(container.__apexChart).forEach((name) => hidden.add(name));
+			// When a shared scope exists it is the source of truth for the tab
+			// group; merging the outgoing chart's own hidden set would re-hide
+			// series the user un-hid elsewhere. Only fall back to it when there
+			// is no shared selection.
+			if (sharedHidden === null) {
+				hiddenSeriesNames(container.__apexChart).forEach((name) => hidden.add(name));
+			}
 			container.__apexChart.destroy();
 			container.__apexChart = null;
 		}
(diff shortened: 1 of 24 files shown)


diyor28:
Dependency of iota-uz/eai#3270 (profitability dashboard). Merge this to SDK `main` first; eai then re-pins `back/go.mod` to the merged commit.

PR #841, opened 2026-06-24
feat(export): GET-download mode for ExportDropdown with busy loader

diyor28:
## What

Adds an opt-in **GET-download mode** to the shared `ExportDropdown` component, for export handlers that **stream the file directly in the response body on a GET route** — as opposed to the existing htmx-POST → build → upload-to-MinIO → `HX-Redirect` mode.

### Why
The single htmx-POST design doesn't fit handlers that stream a file on a GET route (e.g. EAI's portfolio policies/admissions exports): the `hx-post` 404s against a GET route, and an `hx-swap="none"` htmx response can't save a binary body to disk. Consumers worked around it with brittle per-page Alpine overrides that race htmx's own click listener.

### Behavior (`Download: true`)
- Click navigates a hidden `<iframe>` to `ExportURL` with the current filters, a unique `download_token`, and the chosen `format` (a real GET download — no htmx).
- Trigger shows a busy spinner; a light `"preparing"` overlay (`Export.Preparing`) is shown until the handler **echoes `download_token` back as a cookie** (signalling the download has started). A 10-minute deadline always clears the busy state.
- `ParamsFormID` (optional): serialise a filter `<form>` into the export query when filters live outside the page URL. `sort`/`order` are merged from the URL; `page`/`limit`/`per_page` are dropped (export = all rows).

### Backward compatibility
- `Download` defaults to **false** → the existing htmx-POST behavior is byte-for-byte unchanged (guarded by a test). CRM clients / OSAGO / NAPP exports are unaffected.

### Tests
- `TestExportDropdown_DownloadMode` — GET-download markup, no `hx-post`, token-cookie loader wiring, per-format triggers.
- `TestExportDropdown_LegacyHtmxMode` — default mode still emits `hx-post` / `hx-swap="none"`.

Adds `Export.Preparing` to core locales (en/ru/uz/uz-Cyrl/zh).

Consumed by iota-uz/eai (portfolio export 404 fix); EAI pins this via the interim SDK lineage until back/ migrates to the descriptor-nav change on main.

## Summary by CodeRabbit

* **New Features**
  * Added a direct download flow for export with a visible “Preparing your file…” overlay and loading spinner.
  * Export actions now support building export parameters from the current page query (and, when configured, from a selected filter form) instead of the legacy request approach.
* **Bug Fixes**
  * Export format buttons are disabled during file preparation and the UI reliably clears the preparing state when the download starts.
* **Documentation**
  * Updated export localization in multiple languages to include the new “Preparing” message.
* **Tests**
  * Added rendering tests for both the new download mode and the legacy behavior.

### DIFF
--- a/components/export/export_dropdown.templ
+++ b/components/export/export_dropdown.templ
@@ -1,8 +1,10 @@
 package export
 
 import (
+	"fmt"
 	icons "github.com/iota-uz/icons/phosphor"
 	"github.com/iota-uz/iota-sdk/components/base/button"
+	"github.com/iota-uz/iota-sdk/components/loaders"
 	"github.com/iota-uz/iota-sdk/pkg/composables"
 	"net/http"
 )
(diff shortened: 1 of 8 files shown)


diyor28:
Addressed all review feedback in the latest push (`350568d`):

1. **Cookie parsing robustness** (Copilot) — now `document.cookie.split(';').some(c => c.trim() === 'download_token=' + token)`, so it tolerates the optional space after `;`.
2. **Brittle test assertions** (Copilot) — replaced the hardcoded `&#39;` checks with `assert.Regexp(runExport\(\S*excel …)`, asserting the format arg is present without pinning the HTML-entity encoding.
3. **Invalid `ParamsFormID` silent fallback** (CodeRabbit, major) — now fails loud: if `ParamsFormID` is set but the element is missing/not a form, it `console.error`s and returns instead of falling back to `window.location.search` (which for form-only filters could export a broader dataset).
4. **Preserve existing `ExportURL` query** (CodeRabbit, major) — fixed, but **not** via `new URL().searchParams.set`: `.set` would collapse multi-valued filter keys (the chip-builder emits repeated `f=` params), silently dropping filters. Instead used a separator + `params.toString()` which preserves both any pre-existing `ExportURL` query and repeated keys:
   ```js
   const sep = this.$root.dataset.exportUrl.indexOf('?') === -1 ? '?' : '&';
   const target = this.$root.dataset.exportUrl + sep + params.toString();
   ```

`go test ./components/export/...` green; templ/gofmt clean.

PR #850, opened 2026-06-27
feat(lens): add faceted cube filters

### LINKED ISSUE #849: Lens cube: faceted multi-value filters + decoupled "Group by" (replace single-value hierarchical drill)
## Problem

The Lens cube dashboard (e.g. `/insurance/sales-report`) presents dimension chips (Product, Region, Payment, Source, Gender, Vehicle, Age) that users read as **filters/facets**, but they actually drive a **single-value hierarchical drill**:

- `DrillContext` is an *ordered* one-value-per-dimension chain (`pkg/lens/cube/drill_context.go`); `WithFilter` *truncates* everything after the touched dimension.
- A dimension pill doubles as **both** the next-drill target **and** the chart breakdown axis (`pkg/lens/render/templ/dashboard.templ`, the `Remaining` row).
- Filtering is only reachable as a *side effect of clicking a chart bar*; the applied filter then surfaces in a visually separate breadcrumb row, so cause→effect is invisible.

Result: the control conflates "group the charts by X" with "narrow everything to X = value," which mismatches the faceted-search mental model users expect. Multiple users expect these chips to behave as filters.

## Proposed solution

Split the two concerns and make filtering a first-class, multi-value, direct action.

### 1. Cube model (`pkg/lens/cube`)
- Filters become **multi-value & order-independent**: `_f=dim:v1,v2` → OR within a dimension, AND across dimensions. Removing one dimension leaves the others intact (no truncation/hierarchy).
- New `_groupby=<dim>` param drives the chart breakdown, **decoupled** from filters.
- `DrillContext` gains `GroupBy string`; filters keyed by dimension → `[]string`. Legacy `_f` chains still parse (bookmark/back-compat).

### 2. Faceted value lists (new capability)
- Per-dimension distinct `(value, label, count)` honoring the *other* active filters (true faceting). SQL mode: `SELECT col, label, COUNT(...) … WHERE <other filters + date> GROUP BY …`; dataset mode derives from the frame.
- Surfaced on `DrillMeta` so rendering is data-driven.
- **Loaded lazily on dropdown open** via a small HTMX fragment route (zero upfront query cost).

### 3. Rendering (`pkg/lens/render/templ`)
- **Facet dropdown bar**: one dropdown per dimension — checkbox list + counts, search box for high-cardinality dims (Region/Vehicle), **Apply / Clear** per dropdown (batched apply, one request).
- **Applied-filter chips** row: removable `×`, independent of each other.
- **Group by** segmented control (dropdown if many dims) → sets `_groupby`.
- Chart-bar click → *toggles* that value into the dimension's filter set (keeps direct manipulation) instead of re-pivoting.

### 4. Consumer rollout (all cube dashboards, one epic)
- `eai`: `insurance/sales-report` + `crm/summary_report` controllers, `lenssupport`, sales-report templ, `main.css`, locales ×4.
- `eai`: `analytics/drill_service` and any other `cube.Resolve` consumers migrated to the new model.
- Leaf/portfolio passthrough is already multi-value (`portfolioFilterConditionsWithOverrideValues`, `OpIs`, `[]string`) — minimal change.

## Confirmed decisions
- **Facet options:** lazy on dropdown open (HTMX fragment + counts).
- **Apply timing:** explicit Apply button per dropdown (batched).
- **Scope:** all cube dashboards in one epic (engine + sales-report + CRM + analytics drill).

## Risks / constraints
- **Perf** — faceted counts: mitigated by lazy per-dropdown load; cap + search on high-cardinality dims.
- **Back-compat** — every existing `cube.Resolve` consumer must keep working; migrate all in-epic or keep legacy drill API behind the new model.
- **Tailwind scan gap** — the eai CSS build does not scan SDK `pkg/lens`; new dropdown classes need SDK marker classes + raw CSS in eai `main.css` (same pattern as the recent sidebar fix).
- **Multi-repo** — paired PRs in `iota-uz/iota-sdk` + `iota-uz/eai`; submodule bump.

## Phased delivery
1. **Engine** — multi-value/non-hierarchical filters + `_groupby` + faceted resolver (SQL + dataset) + unit tests; legacy `_f` parsing retained.
2. **Templ** — facet dropdowns + applied chips + group-by control + lazy options fragment route.
3. **eai sales-report** — wiring + i18n + CSS → browser-verify both base and drilled states.
4. **CRM + analytics-drill parity** — migrate remaining consumers.

## Acceptance criteria
- [ ] Each dimension is a dropdown facet; multi-select with Apply narrows results (AND across dims, OR within a dim).
- [ ] Applied filters render as independent removable chips; removing one doesn't reset others.
- [ ] "Group by" control changes only the chart breakdown, independent of filters.
- [ ] Facet options + counts load lazily on open, faceted by other active filters.
- [ ] Existing `_f` drill URLs still resolve; all current cube dashboards render unbroken.
- [ ] Leaf "Open in portfolio" carries the full multi-value filter set.
- [ ] i18n keys in all 4 locales; `go vet`, `templ generate`, and tests pass.

---
_EAI is the first consumer; rollout covers all cube dashboards (sales-report, CRM summary, analytics drill). Paired PRs in `iota-uz/iota-sdk` + `iota-uz/eai`._

diyor28:
## Summary
- replace ordered cube drill filtering with multi-value facet filters and `_groupby`
- add lazy facet option resolution for SQL and dataset cubes
- update Lens templ/apex rendering so facet chips, group-by controls, and chart clicks toggle filters

Closes #849

## Testing
- go test ./pkg/lens/cube ./pkg/lens/transform ./pkg/lens/render/templ ./pkg/lens/render/apex ./pkg/lens/compile
- go test ./... (fails only in pkg/middleware because local database `iota_erp` is missing)

## Summary by CodeRabbit

* **New Features**
  * Added grouped “group-by” drill navigation and updated the remaining-facet UI to match it.
  * Enabled multi-value drill filters per dimension, including updated URL/query behavior and toggle handling.
  * Added dataset- and SQL-backed resolved facet options with live-search, lazy HTMX loading, and clear/apply controls.

* **Bug Fixes**
  * Fixed facet resolution and filtering to correctly reflect multi-selected value sets (IN/ANY semantics) and persist selections across navigation.
  * Updated drill action URL behavior for facet toggle interactions.

* **Tests**
  * Expanded drill-context, facet-options, and UI rendering tests for the new multi-value/group-by behavior.

### DIFF
--- a/components/sidebar/sidebar.templ
+++ b/components/sidebar/sidebar.templ
@@ -379,7 +379,6 @@ templ CollapsedGroupFlyout(node NavNode, depth int, navID string) {
 			data-sidebar-nav-id={ navID }
 			data-group-id={ node.ID }
 			data-depth={ depthAttr(depth) }
-			x-transition
 			class="sidebar-flyout fixed z-[70] w-64 rounded-lg p-1.5"
 		>
 			<ul class="flex flex-col gap-1">
--- a/components/sidebar/sidebar_templ.go
+++ b/components/sidebar/sidebar_templ.go
@@ -1178,7 +1178,7 @@ func CollapsedGroupFlyout(node NavNode, depth int, navID string) templ.Component
 		if templ_7745c5c3_Err != nil {
 			return templ_7745c5c3_Err
 		}
-		templ_7745c5c3_Err = templruntime.WriteString(templ_7745c5c3_Buffer, 70, "\" x-transition class=\"sidebar-flyout fixed z-[70] w-64 rounded-lg p-1.5\"><ul class=\"flex flex-col gap-1\">")
+		templ_7745c5c3_Err = templruntime.WriteString(templ_7745c5c3_Buffer, 70, "\" class=\"sidebar-flyout fixed z-[70] w-64 rounded-lg p-1.5\"><ul class=\"flex flex-col gap-1\">")
 		if templ_7745c5c3_Err != nil {
 			return templ_7745c5c3_Err
 		}
@@ -1228,7 +1228,7 @@ func SidebarFallbackIcon(text string) templ.Component {
 		var templ_7745c5c3_Var47 string
 		templ_7745c5c3_Var47, templ_7745c5c3_Err = templ.JoinStringErrs(sidebarFallbackInitial(text))
 		if templ_7745c5c3_Err != nil {
-			return templ.Error{Err: templ_7745c5c3_Err, FileName: `components/sidebar/sidebar.templ`, Line: 403, Col: 32}
+			return templ.Error{Err: templ_7745c5c3_Err, FileName: `components/sidebar/sidebar.templ`, Line: 402, Col: 32}
 		}
 		_, templ_7745c5c3_Err = templ_7745c5c3_Buffer.WriteString(templ.EscapeString(templ_7745c5c3_Var47))
 		if templ_7745c5c3_Err != nil {
--- a/modules/core/presentation/assets/js/alpine.js
+++ b/modules/core/presentation/assets/js/alpine.js
@@ -1464,9 +1464,15 @@ let sidebarCollapsedMenus = () => ({
     const navKey = this.navKeyForElement(el);
     const groupId = el.dataset.groupId;
     const depth = Number(el.dataset.depth || 0);
+    // Read the reactive store map unconditionally so Alpine tracks `menusByNav`
+    // as a dependency even on the first evaluation — when a teleported flyout's
+    // x-bind dataset attributes are not applied yet and navKey is still empty.
+    // Without this, the early return below skips the reactive read and x-show
+    // never re-evaluates after the store updates (collapsed flyouts never open).
+    const menusForNav = this.menusByNav[navKey];
     if (!navKey || !groupId || Number.isNaN(depth)) return null;
 
-    const menu = this.menusByNav[navKey]?.[depth];
+    const menu = menusForNav?.[depth];
     if (!menu || menu.id !== groupId) return null;
 
     return menu;
--- a/pkg/lens/CLAUDE.md
+++ b/pkg/lens/CLAUDE.md
@@ -143,6 +143,8 @@ That means an override query can safely reference params such as:
 - `@f_product`
 - `@f_region`
 
+Drill-filter params (`@f_<dimension>`) are bound as Postgres text arrays so multi-select drills work, so override SQL must use array semantics — e.g. `WHERE pr.id::text = ANY(@f_product)` rather than `= @f_product`.
+
 If a dimension can be drilled further, still give it a valid `.Column(...)` filter expression so later drill levels and KPI stats can apply that filter outside the override dataset.
 
 ## Drill URL Format
--- a/pkg/lens/cube/drill_apply.go
+++ b/pkg/lens/cube/drill_apply.go
@@ -26,7 +26,9 @@ func ApplyDrillFilters(spec CubeSpec, ctx DrillContext, appliers map[string]Dril
 			}).Warn("cube: drill filter has no applier — dimension will be ignored in detail query")
 			continue
 		}
-		applier(filter.Value)
+		for _, value := range filter.values() {
+			applier(value)
+		}
 	}
 }
 
@@ -37,7 +39,9 @@ func ApplyDrillFilters(spec CubeSpec, ctx DrillContext, appliers map[string]Dril
 func ApplyDrill(ctx DrillContext, appliers map[string]DrillApplier) {
 	for _, filter := range ctx.Filters {
 		if applier, ok := appliers[filter.Dimension]; ok {
-			applier(filter.Value)
+			for _, value := range filter.values() {
+				applier(value)
+			}
 		}
 	}
 }
--- a/pkg/lens/cube/drill_context.go
+++ b/pkg/lens/cube/drill_context.go
@@ -10,16 +10,19 @@ import (
 const (
 	QueryFilter    = "_f"
 	QueryDimension = "_dim"
+	QueryGroupBy   = "_groupby"
 )
 
 type DimensionFilter struct {
 	Dimension string
 	Value     string
+	Values    []string
 }
 
 type DrillContext struct {
 	Filters         []DimensionFilter
 	ActiveDimension string
+	GroupBy         string
 }
 
 type Breadcrumb struct {
@@ -42,25 +45,30 @@ func ParseDrillContext(values url.Values) DrillContext {
 		if !ok || dimension == "" || value == "" {
 			continue
 		}
-		ctx.Filters = append(ctx.Filters, DimensionFilter{
-			Dimension: dimension,
-			Value:     value,
-		})
+		for _, item := range strings.Split(value, ",") {
+			ctx = ctx.withFilterValue(dimension, item)
+		}
 	}
 	ctx.ActiveDimension = strings.TrimSpace(values.Get(QueryDimension))
+	ctx.GroupBy = strings.TrimSpace(values.Get(QueryGroupBy))
+	if ctx.GroupBy == "" {
+		ctx.GroupBy = ctx.ActiveDimension
+	}
 	return ctx
 }
 
 func (c DrillContext) Encode() url.Values {
 	values := url.Values{}
 	for _, filter := range c.Filters {
-		if strings.TrimSpace(filter.Dimension) == "" || strings.TrimSpace(filter.Value) == "" {
+		dimension := strings.TrimSpace(filter.Dimension)
+		filterValues := filter.values()
+		if dimension == "" || len(filterValues) == 0 {
 			continue
 		}
-		values.Add(QueryFilter, filter.Dimension+":"+filter.Value)
+		values.Add(QueryFilter, dimension+":"+strings.Join(filterValues, ","))
 	}
-	if trimmed := strings.TrimSpace(c.ActiveDimension); trimmed != "" {
-		values.Set(QueryDimension, trimmed)
+	if trimmed := c.normalizedGroupBy(); trimmed != "" {
+		values.Set(QueryGroupBy, trimmed)
 	}
 	return values
 }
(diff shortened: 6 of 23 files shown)


diyor28:
## QA + review findings (local, Chrome MCP, sdk `d26e9dbc4` + eai tip)

Engine correctness reviewed and the core logic holds — multi-value OR-within / AND-across (SQL `col = ANY(@f_<dim>)` per dim, ANDed; dataset `in` predicates), `_groupby` decoupling, removal-independence (`WithoutDimension`/`ToggleFilter`), legacy single-value `_f` parse, parameterized facet/search values (injection-safe), and faceted counts excluding the self-dimension (`WithoutDimension` before count). Tests/vet pass. Issues:

### 🟠 Facet search "as you type" never fires (live keyup is dead)
- In `FacetOptionsFragment` the search `<input>` has `hx-trigger="keyup changed delay:350ms, search"` but **no request verb of its own**; `hx-get` lives on the parent `<form>`. Under htmx 2.0.2 the input's keyup never issues a request.
- **Verified in-browser**: `htmx.trigger(input,'keyup')` and a native keyup both fire **zero** requests; `htmx.trigger(form,'submit')` **does** fire (`?_facet=region&_facet_search=…`). Server-side search is correct (`_facet_search=Самар` → exactly "Самаркандская область").
- **Net UX**: typing in the facet search box does nothing — only **Enter** (native form submit) filters. The headline "search as you type" doesn't work.
- **Fix**: put `hx-get` + `hx-target` + `hx-include` on the `<input>` itself (classic active-search pattern), or move the `keyup`/`search` trigger onto the form via `hx-trigger="keyup changed delay:350ms from:input[name='_facet_search']"`.

### 🟡 Minor UX
- Facet `<details>` dropdowns don't close each other — opening Регион while Продукт is open leaves both open and overlapping.
- After a search submit the term isn't echoed back into the input (renders empty although the list stays filtered).

### From static review (still applies)
- **Test gap** — the new filter semantics have no direct unit tests: `ParseDrillContext` (multi-value `,` + legacy single-value round-trip), `sqlWhere`/`sqlParams` OR/AND generation, `ResolveFacetOptions` faceted-count-excludes-self, `ToggleFilter` removal-independence. Central regression risk for a `breaking-change` PR — recommend covering before merge.
- **`IsLeaf()` hardcoded to `false`** → `compile.LeafURL` always returns no URL; the cube's `Leaf("/…")` config and the consumers' `isLeaf` redirect branches are now dead. Decide: intentional (faceted model has no leaf auto-redirect) → strip the leaf wiring; or regression → restore real leaf detection.
- **Dead/legacy drill scaffolding** left in place: `WithFilter`/`PopTo`/`Breadcrumbs`/`ContainsDimension`, `resolveLeaf`, `reorderByActiveDimension`, and the unused `drillNavModel` `Current*`/`Trail`/`UpLabel` fields. Mark `Deprecated:` or remove.
- **`RemainingDimensions(spec)` ignores its receiver** and returns all dimensions — misnamed now; rename/drop the receiver.
- `_f` encode logic is duplicated across `Encode`/`WithValues`/`drillURL`/`dimensionTabURL`/`facetOptionsURL` (~5 sites) — extract one helper to avoid format drift.

### ⚠️ Cross-repo: the most serious bug is on the EAI side
- Applying a facet **drops the external date range** (`ActualRangeStart`/`ActualRangeEnd` not included in facet requests) → the whole report recomputes over all-time data (3 368 → 382 949 policies). Details + repro: iota-uz/eai#3314. If the intended contract is that the engine should round-trip *all* non-`_f` query state through facet/group-by URLs, that may warrant an SDK-side fix too (the encode helpers currently forward only `_f`/`_groupby`).

diyor28:
## QA round 2 — more results

### `IsLeaf()→false` is a user-facing regression, confirmed
- The leaf routes still work when hit directly (`/insurance/sales-report/drill/{contracts,policies}` and `/crm/reports/sales/drill/policies` all return **200 with record tables**), but **nothing in the faceted UI links to them** — with `IsLeaf` hardcoded `false`, `compile.LeafURL` never returns a URL and no auto-redirect fires, so the raw-records / "Open in portfolio" drill-down is **orphaned** (reachable only by hand-typed URL). Decide: intentional (then remove the dead `Leaf(...)`/`resolveLeaf`/`isLeaf` wiring **and** the now-unreachable leaf routes/handlers) or a regression (restore real leaf detection + a UI affordance).

### Engine SQL verified correct in-browser (server-side, date present)
- Multi-value OR-within-dimension is exact: `age_group` override `= ANY(@f_age_group::text[])` → 1 band 505, 2 bands **1007 = 505+502**. All dimensions apply with 200, no SQL/runtime errors. The injection-safe param binding and `WithoutDimension` faceted counts hold up under real data.

### Cross-repo: the apex chart-click is the correct reference for the date-drop fix
- The chart-click drill action (apex.go) **does** forward `ActualRangeStart`/`ActualRangeEnd` (period-correct result 3127), whereas the facet apply form / group-by links / facet-option requests built by the templ+encode helpers **don't** — they forward only `_f`/`_groupby`. If the engine's contract is to round-trip *all* non-`_f` query state through facet/group-by URLs, the `_f` encode helpers (`Encode`/`WithValues`/`drillURL`/`facetOptionsURL`) need to carry through passthrough params too. Repro + impact: iota-uz/eai#3314. (Minor: the chart-click action appends the date params twice.)

Search-keyup wiring + the minor UX items from the first comment still stand.

diyor28:
## Questions for the author (@codex)

Three design decisions came out of QA — your call on each before I suggest concrete fixes:

**1. Should the engine round-trip *all* non-`_f`/`_groupby` query params through facet & group-by URLs?**
The apex chart-click drill action forwards `ActualRangeStart`/`ActualRangeEnd` (so chart-click results are date-correct), but the facet encode helpers (`Encode` / `WithValues` / `drillURL` / `facetOptionsURL`) forward only `_f` + `_groupby` — so facet Apply / group-by / facet-option requests drop the date range (this is the eai#3314 blocker: 3 368 → 382 949 policies).
- Was the chart-click carrying the date deliberate, and the facet paths just missed it?
- Should the engine preserve **arbitrary passthrough params** (date + any future external filter) automatically, so every consumer gets it for free? Or is preserving external filters intentionally the consumer's job (hx-include the filter form)? This decides whether the fix lands here or in eai.

**2. Is `IsLeaf()` hardcoded to `false` intentional?**
Leaf routes still return 200 (`/insurance/sales-report/drill/{contracts,policies}`, `/crm/reports/sales/drill/policies`) but nothing in the faceted UI links to them, and `compile.LeafURL` never fires. So the raw-records / "Open in portfolio" drill-down is orphaned.
- Intentional (faceted model has no leaf drill-through)? → then should the dead `Leaf(...)` / `resolveLeaf` / `isLeaf` controller branches **and** the now-unreachable `/drill/*` routes be removed?
- Or a regression to restore (leaf detection + a UI affordance)?

**3. Was facet search meant to be live / as-you-type?**
The search `<input>` has `hx-trigger="keyup changed delay:350ms, search"`, which reads as intended live search — but `hx-get` is on the parent `<form>`, so the input never issues a request; only Enter (native submit) works. Confirm the intent is live-search (→ move `hx-get`+`hx-target`+`hx-include` onto the input, active-search pattern) vs Enter-only (→ drop the keyup trigger).

diyor28:
Answered from the implementation context in the Codex session and pushed follow-up `8261f0b1e`.

1. The engine should own passthrough for Lens-generated facet/group-by requests. This is now handled by carrying the existing request query through `DrillContext.WithValues` and by wiring the async toolbar to include the configured filter form, matching the chart-click behavior. `_facet` / `_facet_search` are stripped from generated drill URLs so lazy-fragment state does not leak.
2. `IsLeaf() == false` was not intentional. That was a regression from moving to facets. `IsLeaf` now returns true when all cube dimensions have an active filter, so existing `Leaf(...)`, `compile.LeafURL`, and consumer `/drill/*` routes stay valid.
3. Facet search was intended to be live/as-you-type. The search input now owns `hx-get`/`hx-target`/`hx-include`, and the fragment keeps the submitted search value populated.

Also addressed the review findings around repeated `_f` encoding, values containing commas, `ParamVariable` preservation in SQL facet params, override-backed SQL facet options, default-dimension trimming, and added focused regression tests.

diyor28:
Follow-up from browser QA: I found that input-level facet search was firing but only sent `_facet_search` plus external date params, not `_facet` / `_groupby`. Fixed in `41e0068e5` by making the search input include its closest facet form plus the external filter selector. Verified with SDK tests: `go test ./pkg/lens/cube ./pkg/lens/render/templ ./pkg/lens/render/apex` and `go vet ./pkg/lens/cube ./pkg/lens/render/templ ./pkg/lens/render/apex`.

diyor28:
## ✅ Re-verified against the fixes (eai `ed770402d` + sdk `41e0068e5`, fresh local build, Chrome MCP)

All four findings confirmed fixed on `/insurance/sales-report` (period 29 May–27 Jun):

| Finding | Before | After |
|---|---|---|
| **Date dropped on facet Apply** (blocker) | 382 949 policies (all-time) | **3 353** — Apply request now carries `ActualRangeStart`/`ActualRangeEnd` alongside the two `_f=region:` params ✅ |
| **Facet counts ignored date** | Город Ташкент 275 778 | **3 352** (period-correct; counts sum to ~3 368) ✅ |
| **Live "search as you type" dead** | keyup fired nothing | typing "Таш" fires `?…&_facet=region&_groupby=product&_facet_search=Таш&ActualRange…` and filters live to Город Ташкент + Ташкентская область ✅ (input now owns `hx-get`/`hx-target`/`hx-include`) |
| **Leaf orphaned (`IsLeaf()→false`)** | no UI path to records | filtering **all 8 dimensions redirects to `/drill/contracts`** (records table, 200) ✅ |

Also re-checked:
- **Group-by + chip removal** still carry the date (period-correct results).
- **CRM `/crm/reports/sales`**: facet counts now respond to an explicit narrow date window (no longer relying on the 30-day fallback) — the `WithActualRange` include path works for dataset mode too.
- `node build-tailwind-css`, `go build`, server boot all clean.

**One cosmetic nit remains**: facet/chart-click URLs duplicate the date params (e.g. `ActualRangeStart` twice). Harmless — flagging only for tidiness.

Nice turnaround — from my side this resolves the blocker + the three majors. 👍

diyor28:
Pushed `60118c9dc` addressing the review feedback:

**Fixed**
- **Positional `GROUP BY`** in `sqlFacetOptionsQuery` (`GROUP BY 1, 2`) — avoids the SELECT-alias/column ambiguity (Copilot).
- **`activeFilterCount`** now counts *distinct* facet values so duplicate `_f` entries don't overstate the badge (CodeRabbit).
- **`cube.Resolve`** wraps its error returns with `serrors.E(op, …)` for operation context (CodeRabbit).
- **Removed dead leaf-resolution code** (`resolveLeaf`, `resolveDatasetLeafDataset`, `leafDatasetNamePrefix`) — confirmed zero callers. The public `Leaf(...)` builder and `CubeSpec.Leaf` field stay for API compat (Copilot).
- **Test names** → `Test*_Scenario` convention (CodeRabbit).
- **Docs**: noted `@f_<dimension>` override params are bound as Postgres `text[]`, so override SQL must use `= ANY(@f_…)` (Copilot).
- templ-fmt blank line before `lensFacetToggleScript`.

**Already resolved in current code**
- Override-dimension facet options are handled via `sqlOverrideFacetOptionsQuery`.
- The lookup contract is `map[string]lens.ParamValue`, so `ParamVariable` is preserved.
- `DefaultDimension` is looked up after `TrimSpace`.
- Facet live-search includes pending checkbox state via `facetSearchIncludeSelector` (`hx-include`).

**Intentional / deferred**
- *Group-by UI control*: the per-dimension group-by selector was intentionally dropped (commit `25b2ea22f`) in favour of the full dimension-panel grid; `_groupby` still works via URL.
- *Facet `COUNT(*)` vs measure-aware count*: left as `COUNT(*)` — the badge is an approximate hint and deriving a `COUNT(DISTINCT …)` expression from the configured measure is a larger change; tracked as a follow-up rather than blocking this PR.

PR #852, opened 2026-06-30
fix(di): inject interface-contract services into handlers

diyor28:
## Problem

`di.H` handlers could not receive services registered as **interface contracts** — only concrete `*T` pointers. Resolving such a handler failed at runtime with:

```
no provider found for type: services.PaymentAllocationService
```

This blocked EAI's payment-allocation feature, whose service is exposed via an interface.

## Fix

`pkg/di`:
- `serviceProvider.Ok` now matches `reflect.Interface` in addition to `reflect.Ptr`, so interface-typed args resolve through the container.
- `provideValue` checks app-provided context values **before** falling through to `resolveProvider` (fixes ordering so app values win).
- `createHandlerFunc` skips eager validation for interface args (they resolve from the container at call time, not registration time).

## Tests

Added `TestH_InjectsInterfaceServiceFromContainer` — a handler taking an interface service resolves it from the container and serves 200. `go test ./pkg/di/...` passes.

## Context

Paired with EAI branch `codex/payment-allocation-transfer` (payment-allocation transfer feature + bank-receipt sync backfill). The EAI submodule pointer bumps to this commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Interface-based dependencies can now be resolved correctly during request handling.
  * Service resolution now validates returned values and shows an error if a service is missing or incompatible.
  * App-level values are checked earlier in the resolution flow for more reliable injection.

* **Tests**
  * Added coverage for interface injection and error cases when services are nil or not assignable.

### DIFF
--- a/pkg/di/providers.go
+++ b/pkg/di/providers.go
@@ -77,8 +77,8 @@ func (p *userProvider) Provide(t reflect.Type, ctx context.Context) (reflect.Val
 }
 
 func (p *serviceProvider) Ok(t reflect.Type) bool {
-	// Basic check: must be a pointer type for services
-	return t.Kind() == reflect.Ptr
+	// Services may be registered as concrete pointers or interface contracts.
+	return t.Kind() == reflect.Ptr || t.Kind() == reflect.Interface
 }
 
 func (p *serviceProvider) Provide(t reflect.Type, ctx context.Context) (reflect.Value, error) {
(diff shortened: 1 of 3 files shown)


diyor28:
Addressed the `serviceProvider.Provide` panic risk in db5816520: it now returns a descriptive error when the container resolves a nil or non-assignable value instead of blindly wrapping it with `reflect.ValueOf`. Added `TestServiceProvider_Provide_RejectsNilService` and `TestServiceProvider_Provide_RejectsNonAssignableService`.

PR #869, opened 2026-07-07
Users list: convert to scaffold table + filterbuilder chips

diyor28:
## Summary
- Converts the Users list page (`/users`) from the legacy `base.Table` + sidebar role/group checkbox filters to `components/scaffold/table` (sfui) with `components/scaffold/filterbuilder` chip filters (role, group, and createdAt folded into one bar).
- Adopts `table.ContentHTMX` as the single HTMX render entry point, replacing the old hand-rolled 3-way render branch.
- Breaking (intentional, no bookmarked links depend on it): URL filter params change from `?roleID=&groupID=&CreatedAt.From=` to the filterbuilder `f=`/`fb=` codec.

## Bug fixes surfaced during the conversion
- `user_query_repository.go`: `scanAndLoadUsers` issued nested relation queries (roles/permissions/groups) while the outer result set cursor was still open. Safe against a pooled connection, but caused `conn busy` errors against a single shared `pgx.Tx` (itf's per-test harness, or any real caller-supplied transaction). Fixed by fully draining the outer cursor into a slice before loading any relation.
- `htmx-alpine-init.js`: the `htmx:load` listener also fired for `document.body` during htmx's own initial page-load `htmx.process()` call, double-initializing Alpine on an already-initialized page and duplicating every `x-for`-rendered node (e.g. two entries per combobox option, site-wide). Fixed by skipping that redundant `document.body` case.

## Test plan
- [x] `go vet ./...`
- [x] `templ generate && just css`
- [x] `go test ./modules/core/...` (all green, including new tests for filter registry/decode/apply, table config action-gating, and a real-DB-row regression test for the conn-busy fix)
- [x] Manual browser verification: add/remove role, group, createdAt chips; infinite scroll; "New user" button permission gating; realtime create/update/delete OOB rows; duplicate-combobox-option bug confirmed fixed on both `/users` and `/users/new`

https://claude.ai/code/session_01J7vskTSCGXfQcLByExgWVr

## Summary by CodeRabbit

* **Documentation**
  * Updated Portuguese (Brazil) Help Center translation terminology.

### DIFF
--- a/modules/core/presentation/templates/pages/users/users_test.go
+++ b/modules/core/presentation/templates/pages/users/users_test.go
@@ -158,7 +158,6 @@ func TestBuildTableConfig_ActionGating(t *testing.T) {
 
 	email, err := internet.NewEmail("action-gating@example.com")
 	require.NoError(t, err)
-
 	noPermUser := user.New("No", "Perm", email, user.UILanguageEN)
 	cfgNoPerm := BuildTableConfig(composables.WithUser(usersTestCtx(), noPermUser), props, req)
 	require.Len(t, cfgNoPerm.Rows, 1)


diyor28:
Addressed the review feedback and the CI failure:

**Review comments**
- Consolidated the role/group filter blocks in `applyUserFilterSet` into a shared `appendInFilter` helper (CodeRabbit).
- Fixed the stale "single-select" doc comment on `buildUserFilterRegistry` — role/group are multi-select (Copilot).
- Rewrote `TestBuildTableConfig_ActionGating` to actually exercise both permission branches with an explicit user in context. Turns out `composables.CanUser` treats an *absent* user as permitted (fail-open for unauthenticated/internal callers), so the original empty-context test was silently checking the "has permission" branch twice, not "no permission" vs "has permission" (CodeRabbit).
- Added `data-col` to each cell in `UserRow`'s realtime OOB render, matching the attributes `Table()` sets on normal rows, so responsive column-hiding/selectors behave the same for realtime-inserted/updated rows (Copilot).

**CI**
- `E2E Tests (shard 4/4)` was failing on `tests/users/register.spec.ts` — root cause: the scaffold table renders `config.Actions` twice (a `md:hidden` mobile copy and a `hidden md:flex` desktop copy), and the test's `a[href="/users/new"]` locator's `.first()` was grabbing the hidden mobile copy at the 1280×720 desktop viewport. Fixed by scoping the locator to `:visible`, matching the existing convention in `finance/table-wide.spec.ts`. Reproduced and verified fixed locally (`npx playwright test tests/users/ --workers=1`, all 5 tests green).
- `Code Quality & Formatting` is failing on a pre-existing, unrelated translation-key gap (5 modules — billing/helpcenter/hrm/oidc/superadmin — are missing a `pt-BR.json` file entirely). Confirmed this already fails identically on `main`'s own CI, in modules this PR never touches, so it's out of scope here.

Claude-Session: https://claude.ai/code/session_01J7vskTSCGXfQcLByExgWVr

PR #919, opened 2026-07-22
feat(lens): explore overlay polish — segment stats, action hierarchy, anchored card

diyor28:
## What

Reworks the Lens explore drill overlay — the contextual card that appears when a
chart segment is clicked — from a bare floating menu into an anchored statistics
card. Scope is the three features chosen for this pass (segment statistics
header, action hierarchy, visual delivery); the excluded rich perspective-preview
feature is not included.

## Before / After

**Before:** a floating card with the segment name, a close ×, plain pill buttons
under «View this segment as», and a plain «Expand segment» row. No segment value
in the header, no visual link to the clicked slice, every action visually equal,
card unanchored.

**After:**

- **Segment statistics header (A).** A color swatch resolved through the same
  `seriesColorResolver` the plot and legend use (no parallel color path), the
  formatted value in `tabular-nums`, and a secondary «78.2% of {total}» line.
  The value/total come from the drill model; the share is rounded once, matching
  the pie slice label (`slicePercentLabel`), so there is no double-rounding drift.
  A small-caps eyebrow names the segment above its title.
- **Action hierarchy (C).** The action the segment resolves to — expansion, else
  the leaf «Open record» — becomes the single filled brand-token primary button
  (which action fires is unchanged, only its visual rank). A quiet «Copy value»
  control copies the formatted figure via `navigator.clipboard.writeText` with an
  `execCommand` fallback (and a silent no-op if neither is available) and swaps to
  a «Copied» confirmation on the button for ~1.5s. Perspective pills stay
  secondary.
- **Visual delivery (D).** An anchor caret points back at the clicked mark and
  flips to the correct card edge when placement changes near a viewport edge. A
  scale(0.96→1) + fade pop-in runs on open, gated off through
  `isVisualRegression()` and `prefers-reduced-motion`. Elevation/border use
  existing theme tokens only. Interactive rows reveal a stepping chevron on hover;
  ArrowUp/ArrowDown roll focus through every interactive element in the card
  (Esc and focus restore behaviour unchanged).

## New runtime i18n keys — consuming apps must translate these

Registered in `pkg/lens/document/i18n.go` + `RuntimeI18nKeys()` (the sync test
`i18n_test.go` enforces both sides). English defaults ship in the runtime; the
host app (e.g. EAI Granite) must add translations:

- `explore.segmentEyebrow` — "Segment"
- `explore.shareOfTotal` — "{share} of {total}"
- `explore.copyValue` — "Copy value"
- `explore.copied` — "Copied"

## VR determinism

The pop-in is the only new motion; it is decided once at mount and skipped when
`isVisualRegression()` is true or reduced motion is requested, so a screenshot
never lands mid-scale. The caret and header are static and resolve from the same
measured-rect settling machinery the overlay already used. A new directly-rendered
`explore--segment-overlay-statistics--{light,dark}` story gives the statistics
header VR coverage that the level-overlay screenshots did not exercise.

Linux baselines were regenerated via the manual `Test, lint & build` workflow
(`lens_vr_update=true`); the promoted candidates differ only in the
overlay-bearing frames (all level/segment overlay screenshots) plus the two new
stories — every non-overlay frame is byte-identical.

## Tests

- `web/lens`: `pnpm typecheck`, `pnpm lint`, `pnpm test` (275 unit tests incl. new
  coverage for the stats header rounding, copy + rejection fallback, primary
  action ranking, and roving keyboard focus), `pnpm build`, `pnpm ladle:build`.
- `go test ./pkg/lens/document/` — i18n sync green.
- Lens Visual Regression — green against the promoted Linux baselines.

https://claude.ai/code/session_01SpFsbgRTjyBtqBRM17RC9b

## Summary by CodeRabbit

* **New Features**
  * Enhanced Explore drill overlays with segment labels, totals, color accents, and share-of-total information.
  * Added copy-to-clipboard support with visual confirmation and fallback handling.
  * Improved keyboard navigation, overlay positioning, caret indicators, animations, and action styling.
  * Added light and dark segment overlay previews.

* **Tests**
  * Added coverage for copying values, keyboard navigation, overlay content, and light/dark visual stories.


diyor28:
Follow-up fix (live verification on the real profitability dashboard found feature A absent for the primary click):

The overlay opened bare — «SEGMENT» eyebrow + title + perspective pills, but no swatch, value, share-of-total or «Copy value» — for a clicked pie segment. Root cause: `Level.encoding` is optional; a drill level that shares its panel's frame shape declares none, and the chart reads `panel.encoding` for it via `viewPanel`. The overlay model read the level's absent encoding directly, so `rowForNode`/`valueForNode` resolved nothing exactly for those levels (the most common dashboard case). The title still rendered because `labelForNode` returns `node.label` directly; value/share/total/copy are gated on the value the builder never populated.

Fix: thread the effective encoding `level.encoding ?? panel.encoding` through `drillTargetForNode`/`drillTargetForLevel` and the row/label/value helpers, and wire the overlay `valueFormat` off `viewPanel`'s encoding/format so the figure keeps money formatting. New `explore.test.tsx` case drives the real builder path (panel + frame + `openForMark` click) on a document whose drill level declares no encoding, asserting value + share + copy render. VR pixels unchanged (no story exercises a missing-encoding document; local compare clean), so the promoted Linux baselines stand.

Commits: bbff0269a (fix), 008246b2f (dist rebuild).

PR #937, opened 2026-07-27
feat(lens): radial panels, waterfall bridges and a printed audit report

diyor28:
## Summary

Three things, in the order they were built:

1. **Radial panels** — a generic Lens panel with concentric-progress and multi-ring partition modes.
2. **Waterfall** — a bridge tells its whole story inside the plot: split bands, callouts, an underlay for what a deduction leaves behind.
3. **A printed report** — `?lens-print-preview=1` and the print button render the dashboard as a document: cover, contents, numbered chapters, an appendix of detail, and a methodology page. Everything the screen keeps behind a hover, a click or a chip is placed where it works on paper.

## API

- `panel.RadialBars(id, title, dataset, maximum)`, `panel.MultiRingDonut(id, title, dataset, rings...)`, `panel.RadialRing{Key, Label, Order, Total}`, `RadialTolerance(tolerance)`, `panel.RadialNodeKey(ringKey, categoryKey)`
- cascade: `split` / `splitLabel` encoding, `bridgeLayout: waterfall`
- container groups carry a caption; a served frame carries its own presentation and palette; a drill level may mix drillable and terminal rows
- `ChartInput.locale` — a chart writes its own shares in the reader's notation
- ~30 `print.*` i18n keys (`document.RuntimeI18nKeys()` enforces two-way parity with the runtime call sites)

## What the printed report does

- a ratio's calculation prints under its KPI tile as a formula, not as five one-row tables in the appendix
- `confidence` / `availability` print as chips beside the number, are counted on the cover and defined in a glossary
- a cascade's evidence table is built from the stages the plot draws, so the table cannot disagree with the picture
- long quarterly series print as a compact chart plus the last 8 periods, and say so
- truncated tables say what share they kept; empty readings are named rather than printed as "no data" cards
- shares below the column's resolution read `< 0.1 %`; a bridge that tints its stages prints a tone key

## Verification

- `just lens check` — 545 tests
- `just lens build` (embedded dist rebuilt and committed)
- `go test ./pkg/lens/...`
- read as real PDFs (Chromium `page.pdf`) against a production-scale dashboard, page by page, in Russian

## Known limitation

Page numbers are not possible in this path: the export is the browser's own print, and Chrome does not support `@page` margin boxes. Chrome's "Headers and footers" option supplies them.

## Follow-up

Linux VR baselines still need promotion through a `lens_vr_update` CI run; the lane is in bootstrap mode until then.


diyor28:
Added generic browser audit printing for Lens dashboards in `24a8be9c5`.

- builds a flat, non-interactive report from every analytical drill level, perspective, dynamic child and company-level drawer
- renders every chart with an adjacent textual evidence table containing compact values, exact values and shares
- prints through the browser using A4 landscape CSS; no server-side PDF renderer
- keeps the report off-canvas through a body portal until print, so dashboard layout/overflow cannot clip it
- skips per-row/per-sector drawer duplication because those values are already preserved in the evidence table
- uses cancellable, isolated query clients with explicit timeout disclosure

Verification: `pnpm check` (491 tests), bundle boundary, production build, and `go test ./pkg/lens/...`. A real Chrome PDF from profitability produced 95 sections, 50 charts, 380 evidence rows, 69 A4 pages, zero unresolved view-choice screens, and no limitations block.

diyor28:
Review findings addressed in 08e89f0b. Where a finding was real I fixed the cause and pinned it with a test; where it was not, the reason is below.

**Fixed**

- *Partition shares grouped by denominator* — two rings both declaring 100 were reconciled as one decomposition. Grouping is by ring key now; `panels.test.tsx` fails on the old behaviour (each ring's thirds print 33.4/33.3/33.3 instead of an unreconciled 33.3 ×6).
- *Printed shares divided by the row sum* — the table now divides by the ring's declared total, the same whole the chart uses.
- *`levelTotal` summed every ring* — a partition radial returns no level total; one headline was being counted once per ring.
- *Legend colour pins drifted after a toggle* — pins are positional, so they are filtered with the rows they pin. Test added.
- *`RadialNodeKey` HTML-escaped `<`, `>`, `&`* — the runtime builds the key with `JSON.stringify`, which does not; a category named `A&B` had two identities. `SetEscapeHTML(false)` plus a table test against the `JSON.stringify` form.
- *`validateRadialConfig` could index -1* — `validateEncodingFields` rejects a missing column first, so it was unreachable, but a guard now keeps a reordering of the two from turning a bad document into a panic. Both orders are pinned by tests.
- *Radial tooltip used `\n`* — `<br/>`, which the HTML tooltip body honours.
- *Apex radial* — `chartType` branches on the mode: a partition radial draws as a donut instead of a radial bar it cannot represent, with a comment that concentric rings are the React runtime's job. Radial side legends now move below the plot at the tablet breakpoint like pie and donut.
- *Tabs unhidden during print* — inactive tabs still mount so the report can resolve their frames, but stay `hidden` while it builds.
- *`AbortSignal.timeout` / `.any`* — feature-detected with `AbortController` fallbacks.
- *Terse print rejection* — names the dashboard that declares no query endpoint.
- *O(n²) legend de-duplication* — `Set`-based.
- *Radial validation suite* — table-driven.

**Not changed**

- *`RadialConfigSchema` as a discriminated union* — `web/lens/src/contract/schemas.ts` is generated by `cmd/lens-typegen` from the Go contract and cannot be hand-edited. Both invariants are enforced in `pkg/lens/document/validate.go` (a positive finite `max` for progress, a non-empty reconciling `rings` for partition, non-negative tolerance), which is the gate every document passes through before it is served. Expressing them in the emitted Zod would be a typegen change, not a schema edit.
- *Cube dimension radial → bar* — kept, now with the reason in the code: a cube dimension cannot state the maximum or the ring totals a radial needs, and `panelBuilder` has no error path. Radial panels belong in a hand-written `DashboardSpec`.

PR #989, opened 2026-08-05
fix(lens): segment-bar hover parity with pie, and stop leaking series0 to the tooltip

diyor28:
Two defects a reader meets on the same dashboard.

## Hover

A pie sector grows under the pointer — ECharts' default pie `emphasis` scales it — while a segment bar (`lensspec.SegmentBar`, wire kind `coverage`) only dimmed its siblings and brightened the active one. Same page, two chart families, two different answers to "what does hover mean here".

The blocker was the track: `.lens-coverage-track` is a fixed `lens-h-2.5` with `lens-overflow-hidden lens-rounded-full`, so it owns the pill silhouette and clips its children to it. A segment growing on its own would square off the rounded ends.

So **the track grows, not the segment** — and the siblings give up the height it gained (`scaleY(0.625)`), which is what makes the hovered one read as having grown rather than the whole bar merely getting fatter. `transform` joins the transition list; `prefers-reduced-motion` drops both transitions.

The **bullet variant** keeps the colour treatment and forgoes the growth: its target marker hangs at a fixed offset over the track, so a track that changed height would slide out from under its own marker.

## `series0`

The tooltip on a single-series chart printed ECharts' internal `series0` as if it were the reader's own label for the measure.

`showSeriesName` was `Boolean(input.encoding.series)` — but **a declared series role is not a series the frame has**. A progressive dashboard builds its panel shell before any frame exists, so the wire encoding it publishes is the panel's *declaration*, and every panel declares the default field names whether it uses them or not. The name then falls back to whatever ECharts auto-assigned.

It now goes through `availableEncodingField`, so the frame is the authority on what a role resolves to — exactly as it already is for the `category` axis on the line below.

## Tests

- `options.test.ts`: 'ignores a declared series role the frame does not carry'.
- `vr/lens.spec.ts`: 'a hovered coverage segment grows out of its quieted siblings' **measures** the heights rather than photographing them. The growth is the contract; a screenshot baseline of it would need re-blessing every time the story's palette moved.
- `just lens check` green (893 tests).

### Pre-existing VR failures

`just lens vr` fails on darwin baselines unrelated to this change — choropleth, distributions, comparison-interactions. The last differs in page **height** (1203 vs 1339), so it is a layout drift, not an anti-aliasing one. Reproduced on a clean checkout of this branch's base; **not** rebaselined here.

## Downstream

Consumed by iota-uz/eai as part of the `/analytics/profitability` pre-release fixes. The EAI PR will carry `Depends-On: iota-uz/iota-sdk#<this>`.

---
<a href="https://app.blacksmith.sh/iota-uz/codesmith/iota-sdk/pr/989"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788523429&installation_model_id=2042&pr_number=989&repository=iota-uz%2Fiota-sdk&return_to=https%3A%2F%2Fgithub.com%2Fiota-uz%2Fiota-sdk%2Fpull%2F989&signature=d6a20ec7c9870409da675dedf957779933e1226a374f480fe3882abf0706ed01"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed single-series chart tooltips so they no longer display synthetic series names when configured fields are unavailable.
* **Visual Improvements**
  * Improved coverage visualizations by emphasizing active segments and de-emphasizing inactive segments during hover.
* **Accessibility**
  * Disabled coverage animations for users who prefer reduced motion.
* **Tests**
  * Added regression coverage for chart tooltips and interactive coverage highlighting.


diyor28:
Обе заметки отработаны в 5d3231582.

- `options.test.ts` — добавил проверку самой цифры: тест доказывал, что синтетическое имя ушло, но не что значение осталось.
- `lens.spec.ts` — предложение сделать `toBeLessThan` я не принял: соседи гасятся ровно тем коэффициентом, который держит их на исходной толщине (16px × 0.625 = 10px), поэтому строгое «стало тоньше» требовало бы поведения, которого в дизайне нет — тест бы падал. Но замечание по сути верное: утверждение читалось как пустое. Написал, почему оно не пустое (без правила сосед вырастает выше исходного и проверка падает), и добавил относительное утверждение, ради которого анимация и делалась — сосед тоньше сегмента под курсором.

PR #993, opened 2026-08-07
feat(billing): implement Click payment reversal

diyor28:
## Problem

`clickProvider.Refund` and `clickProvider.Cancel` are `panic("implement me")`. Any caller that reaches `BillingService.Refund` for a Click transaction crashes the process, so in practice no product can offer a refund on the gateway that carries the largest share of Uzbek card payments.

## Solution

Implement `Refund` against the documented Merchant API reversal endpoint:

```
DELETE {api}/payment/reversal/{service_id}/{payment_id}
Auth: {merchant_user_id}:{sha1(timestamp + secret_key)}:{timestamp}
```

(https://docs.click.uz/en/merchant-api-request/)

Behaviour that the contract forces:

- **Full reversal only.** The request carries no amount. A `quantity` smaller than the transaction amount is refused with `ErrClickPartialRefund` *before* the call, rather than silently reversing the whole payment.
- **No payment id, no reversal.** Without the Complete callback there is no `click_trans_id` and nothing was captured; refused with `ErrClickPaymentIDMissing`.
- **A refusal is an answer, not an outage.** Click's negative `error_code` (already reversed, previous-month window, card-processor refusal) surfaces as `ClickReversalError{Code, Note}`, so a caller can distinguish it from a transport failure and show the operator why. Only a non-negative `error_code` marks the transaction `refunded`.
- An HTTP error page is neither success nor refusal — it is an error.

`ClickConfig` gains two optional fields, both backward compatible:

- `APIURL` — override the Merchant API base (sandbox, or a test server).
- `HTTPClient` — supply a client wrapping `middleware.LogTransport` so the outbound call lands in the request log.

`Cancel` is left as-is; it is a different operation and out of scope here.

**Payme is deliberately not implemented.** Its Merchant API is inbound-only — `CancelTransaction` is a callback the merchant *implements*, not a method it calls — and a checkout-link payment can only be refunded from the Payme cabinet or via support. `receipts.cancel` belongs to the Subscribe API and is not documented to accept a checkout-link transaction id. Leaving the panic is honest; a fake implementation would not be.

## Verification

`go test ./modules/billing/infrastructure/providers/...` — new `click_provider_test.go` covers, against an `httptest` server:

- the exact method, path and `Auth` header, including recomputing the SHA-1 digest and bounding the timestamp;
- a gateway refusal surfacing as `ClickReversalError` with code `-31007`;
- a partial amount refused without any request reaching the gateway;
- a transaction with no `click_trans_id` refused without a request;
- a 502 HTML page treated as a failure and explicitly *not* as a refusal.

`go vet ./modules/billing/...` clean.

Not exercised against the live Click sandbox — the endpoint contract is asserted from the documentation and the request is asserted literally in tests.

---
<a href="https://app.blacksmith.sh/iota-uz/codesmith/iota-sdk/pr/993"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788706823&installation_model_id=2042&pr_number=993&repository=iota-uz%2Fiota-sdk&return_to=https%3A%2F%2Fgithub.com%2Fiota-uz%2Fiota-sdk%2Fpull%2F993&signature=2fa0b83864eb33db9dac44de1de045722800be032f8ae9e1c52082cf46f25d91"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for full payment refunds through Click’s reversal service.
  * Refund requests now include secure authentication and configurable API connectivity.
  * Gateway refusal details and refund status updates are surfaced clearly.

* **Bug Fixes**
  * Partial refunds and refunds without a completed payment are rejected before submission.
  * Improved handling of malformed or non-standard gateway error responses.

* **Tests**
  * Added coverage for successful refunds, authentication, validation, gateway refusals, and connection failures.

### DIFF
--- a/modules/billing/infrastructure/providers/click_provider.go
+++ b/modules/billing/infrastructure/providers/click_provider.go
@@ -3,12 +3,46 @@ package providers
 
 import (
 	"context"
+	"crypto/sha1"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
+	"net/http"
 	"net/url"
+	"strconv"
+	"time"
 
 	"github.com/iota-uz/iota-sdk/modules/billing/domain/aggregates/billing"
 	"github.com/iota-uz/iota-sdk/modules/billing/domain/aggregates/details"
 )
+
+// DefaultClickAPIURL is the Click Merchant API base used when ClickConfig.APIURL
+// is empty. https://docs.click.uz/en/merchant-api-request/
+const DefaultClickAPIURL = "https://api.click.uz/v2/merchant"
+
+// ErrClickPartialRefund is returned when a refund asks for less than the full
+// transaction amount. The Click reversal endpoint takes no amount: it reverses
+// the whole payment or nothing.
+var ErrClickPartialRefund = errors.New("click supports full reversal only, not partial refund")
+
+// ErrClickPaymentIDMissing is returned when the transaction carries no Click
+// payment identifier, which happens when the payment never reached the Complete
+// callback. Nothing was captured, so there is nothing to reverse.
+var ErrClickPaymentIDMissing = errors.New("click transaction has no payment id to reverse")
+
+// ClickReversalError is a rejection reported by Click itself, carrying the
+// gateway's own code so the caller can tell "already reversed" from "refused by
+// the card processor" instead of collapsing both into a generic failure.
+type ClickReversalError struct {
+	Code int32
+	Note string
+}
+
+func (e ClickReversalError) Error() string {
+	return fmt.Sprintf("click reversal rejected: code=%d note=%s", e.Code, e.Note)
+}
 
 type ClickConfig struct {
 	URL            string
@@ -16,6 +50,12 @@ type ClickConfig struct {
 	SecretKey      string
 	MerchantID     int64
 	MerchantUserID int64
+	// APIURL overrides the Merchant API base (sandbox, or a recording proxy in
+	// tests). Empty means DefaultClickAPIURL.
+	APIURL string
+	// HTTPClient performs Merchant API calls. Supply one wrapping
+	// middleware.LogTransport to get the outbound request into the request log.
+	HTTPClient *http.Client
 }
 
 func NewClickProvider(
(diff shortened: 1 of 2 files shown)


diyor28:
Thanks — one of the five findings was a real defect and is fixed in `f970b9e99`, with reasons for the other four.

**Fixed: quantity greater than the captured amount silently reversed everything.** `quantity < amount` was the wrong guard — `Refund(t, 140_000)` on a 100 000 payment fell through to a full reversal, so a caller could believe it returned money that was never captured. The check is now exact equality in both directions (`ErrClickPartialRefund`, message widened to "reverses the exact captured amount only"), with `TestClickRefundRejectsAnAmountLargerThanWasCaptured` asserting nothing reaches the gateway.

**Not taken — returning the transaction alongside `ClickReversalError`.** The caller is `BillingService.Refund`, which runs inside `composables.InTx` and returns on a provider error without saving; a returned transaction would be discarded. Returning `(nil, err)` also keeps a caller from persisting a transaction Click just refused. The gateway's code and note are already on the error itself, which is the part a caller needs.

**Not taken — wrapping in `serrors.E`.** Nothing in `modules/billing` uses `serrors`; providers return plain errors and the callers wrap. Introducing it in one provider method would make the package inconsistent without changing what a caller can do — `errors.Is`/`errors.As` already work.

**Not taken — renaming tests to `TestRefund_X`.** These names state the behaviour being protected ("rejects an amount larger than was captured"), which is what a future reader needs when the test fails. Both styles exist in this repo.

**Not taken — `assert` instead of `require` for the independent checks.** In `TestClickRefundAddressesTheReversalEndpointAndSignsTheRequest` the assertions are not independent: the `Auth` header is split into three parts and the digest recomputed from part 3, so continuing after a shape mismatch produces an index panic rather than extra information.

Lint failures (4 × errcheck on the test server's `fmt.Fprint`, 1 × testifylint `NotErrorAs`) are fixed in the same commit. `golangci-lint run modules/billing/infrastructure/providers/...` — 0 issues.

PR #996, opened 2026-08-07
Lens: honest cascade gaps, a period picker with room, and failure copy that says what broke

diyor28:
Three Lens changes that surfaced from one live review of the EAI profitability and trends dashboards.

## A cascade stage may have no value

A producer that cannot know a stage's amount sends `null`. `numeric()` coerced that to `0`, and the bridge then invented a movement out of it.

Observed on a real dashboard: a reserve stage annotated «Нет данных» rendered a **−149,00 млрд** deduction — exactly the running total it stood at, dropped to zero — and every delta after it was computed from a total that never existed. Separately, a `final` row with no value missed the `end` branch, so the closing synthesizer appended a *second* closing column; that synthesized item deliberately carries no `rowIndex`, so the duplicate sitting next to the real one was inert while still showing the producer's call-to-action.

Now: `hasValue` (via the existing `finiteNumber` the metric panels already use) instead of coercion; an unknown stage draws as a gap at the last **known** running total with a dash, its label and its annotation; deltas skip it so one hole no longer corrupts the rest of the bridge; an unknown `final` row closes in place and keeps its `rowIndex`; and a synthesized closing inherits the `rowIndex` and annotation of the stage it repeats rather than becoming a dead clone.

Unknown and exactly-zero stay visually distinct — `noMovement` supplies the geometry, `unknown` the dash and the lighter weight.

## The period picker

It had been redesigned three times and still read as cramped. The card was 640px for two months at 32px cells; the selected year was styled like a text input; presets had nothing to scan by; month navigation was a bare caret.

Widened to 860 with 40px cells and real padding. Selected year is a filled pill with a brand edge. Presets carry icons and their groups carry headings. Month carets became bordered buttons. From/To fields got a calendar glyph, the duration line got a badge.

Every month now draws on the same **six-week grid**, so a month no longer ends in a row holding a single date, both panes pad symmetrically, and the popover stops changing height as you page through it. «All time» moved inside the completed group — the rule above it came from `.lens-filter-preset-clear`'s border, not from the model, and it was orphaning the entry.

Two existing wash tokens raised for contrast (also in dark). No new colours; one new Phosphor glyph.

## Failure copy

A timed-out panel and a broken backend rendered the same sentence, so the one actionable thing — the period is too large, shorten it — was the one thing the panel would not say.

The execution error is now classified into a small closed vocabulary (`timeout`, `canceled`, `unknown`), carried on the wire, and used to select the message. An absent or unrecognized reason falls back to the host's own generic sentence, so an older host renders exactly as before and no board ever mixes languages. A cancelled load keeps its box and Retry but stops claiming to be an alert.

The classifier lives in `pkg/lens/document` rather than `pkg/lens/runtime`: `document` already imports `runtime`, so the reverse would be a cycle — and the reason is wire contract anyway.

## Verification

`just lens check` passes (typegen drift, tsc, eslint, stylelint, 950 tests). New stories cover the unknown-stage cascade and the three failure reasons in both themes, registered in the VR manifest.

**VR baselines are deliberately untouched** — `git status web/lens/vr/baselines` is empty and `vr-update` was never run. A local full run reports ~139 failures against stale gitignored darwin PNGs that predate this branch; the new stories have no Linux baseline yet and will take CI's missing-baseline candidate path.

https://claude.ai/code/session_01732RPPEn8jhTdBS6KhNX9k

---
<a href="https://app.blacksmith.sh/iota-uz/codesmith/iota-sdk/pr/996"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788728065&installation_model_id=2042&pr_number=996&repository=iota-uz%2Fiota-sdk&return_to=https%3A%2F%2Fgithub.com%2Fiota-uz%2Fiota-sdk%2Fpull%2F996&signature=891af8e819913f22e2a6df7b0150da43c2dcdd5f7bb438cca69b1819fb07ac58"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Panel errors now show localized timeout, cancellation, or generic messages without exposing technical details.
  - Failure responses include machine-readable classifications.
  - Unknown cascade and waterfall values display as em dashes with accessible “Unavailable” labels.
  - Added icons for time-period controls, including an “All time” indicator.
  - Simplified table footers for small, unpaginated result sets.

- **UI Improvements**
  - Expanded and reorganized the date picker with larger calendars, clearer range controls, and responsive behavior.
  - Calendar months now use consistent six-week layouts.
  - Improved visual distinction and navigation cues for unavailable chart values.


diyor28:
Addressed in eaa3ded63 / ae6290212 / e8b0addb0:

- `failure.go:15` (comment verbosity, trivial) — left as-is. Each flagged comment states a non-obvious invariant (why `Reason` is optional, why `panelFailure` exists beside `panelError`, why the classification happens where it does), not a restatement of the code beneath it — that's the bar this repo holds comments to, and CodeRabbit's own severity tag agrees (`Trivial | Low value`).
- `failure.go:68` (timeout substring over-match) — fixed. Matches driver/worker timeout signatures instead of a bare `"timeout"` substring; added the `"invalid timeout configuration"` regression case.
- `CascadePanel.tsx:558` (underlay drawn under an unknown stage) — fixed. `underlayHeight` now returns `undefined` for `unknown` stages; added a model-level assertion.
- `panelFailure.ts` (explicit English translation indistinguishable from absent) — fixed with a sentinel default, keeping each `translate('panel.error...', ...)` call a source literal so `pkg/lens/document/i18n_test.go`'s call-site scanner still finds it (a first attempt through a small wrapper broke that scan — caught by CI, reverted to this shape). Added the explicit-English-host regression test.
- `lens.spec.ts:280` (missing baselines for new stories) — resolved by the baseline promotion commit; all six are in `vr/baselines/linux` now.
- `TablePanel.tsx` (no coverage for the hidden-footer branch) — added a direct unit assertion (`.lens-table-footer` absent under threshold, present with a row count over it) rather than a dedicated Ladle story; the existing `table-readability` stories already exercise the state visually and are in the promoted VR baselines.

Full lens Go (`go test ./pkg/lens/...`) and web (`vitest run`, 954 tests) suites pass.

PR #1013, opened 2026-08-13
[P1] fix(htmx): encode HX-Trigger payloads as valid ASCII-only JSON

diyor28:
## Problem

Toast details were assembled with `fmt.Sprintf` and written as raw UTF-8 into the
`HX-Trigger` header. Go writes header values as bytes, and browsers decode
response header values as ISO-8859-1 (XHR/fetch), so anything above U+007F is
re-interpreted one byte at a time before htmx ever calls `JSON.parse`. A
production operator saw `Ð\x9De ÑÐ´Ð°Ð»Ð¾ÑÑ Ð¿ÑÐ¾Ð²ÐµÑÑÐ¸ Ð¾Ð¿Ð»Ð°ÑÑ…`
where the server sent «Не удалось провести оплату…». English toasts are
unaffected, which is why this went unnoticed.

Two defects:

1. **Non-ASCII does not survive the header** — every Cyrillic (and Uzbek Cyrillic,
   and any non-Latin) toast in every consumer arrives as mojibake.
2. **The JSON was built by string formatting** — a `"`, `\`, newline or control
   character in a title or message produced an invalid document, htmx's parse
   failed, and the toast silently never appeared. It was also an injection into
   the client's `JSON.parse`.

## Fix

- Toast details are marshalled with `encoding/json` (a struct, so field order is
  deterministic) instead of `fmt.Sprintf`.
- The event name is marshalled as a JSON string rather than concatenated.
- Every rune above U+007F in the assembled header value is escaped as `\uXXXX`,
  with a UTF-16 surrogate pair for runes outside the BMP.

The escaping is applied to `SetTrigger`, `TriggerAfterSwap` and
`TriggerAfterSettle` — not just to the toast helpers — because `detail` is
contractually a raw JSON fragment and several callers pass a payload they
marshalled themselves; `encoding/json` emits non-ASCII as raw UTF-8, so those
triggers carried mojibake by exactly the same route. Escaping the finished
document (rather than re-marshalling it) is safe because non-ASCII cannot legally
appear outside a JSON string literal, and it leaves a caller's key order and
number formatting untouched.

**The client needs no change.** `\uXXXX` is valid JSON and `JSON.parse` restores
the original string, verified end to end against the listener in
`components/base/toast/toast.templ` (`@notify.window="addNotification($event.detail)"`).

Before:

    Hx-Trigger: {"notify": {"variant": "error", "title": "Ошибка", "message": "Не удалось провести оплату \"AB 1234567\""}}

After:

    Hx-Trigger: {"notify":{"variant":"error","title":"Ошибка","message":"Не удалось провести оплату \"AB 1234567\""}}

Note the separators are now compact (`{"notify":{…}}` rather than
`{"notify": {…}}`). That is invisible to `JSON.parse` and to the `Contains`-based
HTMX trigger assertions in `pkg/itf`.

The detail-less form (`SetTrigger(w, "event", "")` → a bare event name, which is
what htmx expects for an event without a payload) is unchanged byte for byte.

## Tests

Table-driven tests in `pkg/htmx` assert on the header **as serialized to the
wire**, not on the Go string in the header map — asserting on the map would pass
on raw UTF-8 while the browser still receives mojibake, so every unicode case
requires each byte to be below `0x80`. Covered: Cyrillic and a non-BMP emoji
through all three toast helpers; `"`, `\`, a newline and a control character;
a pre-marshalled UTF-8 `detail` passed straight to `SetTrigger`; an event name
containing a quote; the empty-detail path (asserted with `Equal`); and an
ASCII-only toast, whose parsed payload is unchanged.

Verified against the pre-fix code with public-API-only equivalents: the Cyrillic,
quote-in-message and pre-marshalled-detail assertions fail, the empty-detail
assertion passes.

## Not in this PR

- `Location` has the same string-concatenation defect (`{"path":"`+path+`", …`)
  and the same raw non-ASCII exposure. Left for a separate change.
- `Redirect`/`PushURL`/`ReplaceURL` pass URLs through unchanged — percent-encoding
  is the caller's responsibility there.
- A non-ASCII *event name* with an empty detail still travels raw, because that
  form is not JSON and escaping it would change the dispatched event name.

---
<a href="https://app.blacksmith.sh/iota-uz/codesmith/iota-sdk/pr/1013"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789212262&installation_model_id=2042&pr_number=1013&repository=iota-uz%2Fiota-sdk&return_to=https%3A%2F%2Fgithub.com%2Fiota-uz%2Fiota-sdk%2Fpull%2F1013&signature=9ba1ab32585eade8088c8731420fce7c9f4c68ff179a127a384718bc6bf8223d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved HTMX event and toast notifications containing non-ASCII characters.
  - Preserved valid JSON formatting for quotes, backslashes, control characters, Unicode, and surrogate pairs.
  - Maintained existing behavior for ASCII content and events without details.
  - Safely normalized preformatted event details while preserving caller-provided data.
  - Ensured notification titles and messages remain intact when containing special characters.

### DIFF
--- a/pkg/htmx/htmx.go
+++ b/pkg/htmx/htmx.go
@@ -106,7 +106,14 @@ func triggerHeaderValue(event, detail string) string {
 	if detail == "" {
 		return event
 	}
-	name, _ := json.Marshal(event) // marshalling a string cannot fail
+	// Marshalling a string cannot fail, but the error is answered rather than
+	// discarded: emitting half a document would hand htmx a payload it cannot
+	// parse, whereas the bare event name is the form it already accepts for a
+	// detail-less event.
+	name, err := json.Marshal(event)
+	if err != nil {
+		return event
+	}
 	return escapeNonASCII(`{` + string(name) + `:` + detail + `}`)
 }
 
@@ -215,11 +222,17 @@ type toastDetail struct {
 // characters in a title or message stay valid JSON instead of breaking the
 // client-side parse.
 func toastPayload(variant ToastVariant, title, message string) string {
-	detail, _ := json.Marshal(toastDetail{ // marshalling strings cannot fail
+	// A struct of strings cannot fail to marshal, but the error is answered
+	// rather than discarded: an empty payload degrades to the bare `notify`
+	// event, while a truncated one would break the client-side parse.
+	detail, err := json.Marshal(toastDetail{
 		Variant: variant,
 		Title:   title,
 		Message: message,
 	})
+	if err != nil {
+		return ""
+	}
 	return string(detail)
 }
 
--- a/pkg/htmx/htmx.go
+++ b/pkg/htmx/htmx.go
@@ -56,18 +56,13 @@ func Reswap(w http.ResponseWriter, swapStyle string) {
 	w.Header().Add("Hx-Reswap", swapStyle)
 }
 
-// escapeNonASCII rewrites every rune above U+007F as a \uXXXX escape sequence.
+// escapeNonASCII rewrites every rune above U+007F as a \uXXXX escape sequence
+// (a surrogate pair outside the BMP, U+FFFD for invalid UTF-8), keeping a header
+// value pure ASCII. Browsers decode header bytes as ISO-8859-1, so raw UTF-8
+// reaches htmx as mojibake, while JSON.parse restores the escaped form intact.
 //
-// Header values are transported as bytes and decoded by browsers as ISO-8859-1
-// (XHR/fetch), so raw UTF-8 reaches htmx as mojibake. \uXXXX is valid JSON and
-// JSON.parse restores the original string, so the payload survives intact while
-// the header itself stays pure ASCII. Runes outside the BMP are emitted as a
-// UTF-16 surrogate pair, as JSON requires.
-//
-// Escaping is applied to an assembled JSON document rather than to its parts:
-// non-ASCII cannot legally appear outside a JSON string literal, so this cannot
-// corrupt a valid document, and a caller-supplied payload is never re-parsed or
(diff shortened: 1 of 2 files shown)


diyor28:
Both comment threads addressed in b993b79.

- `escapeNonASCII` keeps its purpose and output behaviour (surrogate pairs, U+FFFD) plus the one sentence a reader cannot derive from the body — that browsers decode header bytes as ISO-8859-1 — and drops the rest.
- The two marshal comments now state the fallback contract, as proposed.
- The seven test doc comments lose their second paragraph; where the "falsely green" note recorded a real trap (requireASCII is the load-bearing assertion, Equal rather than Contains, the caller's document is never re-marshalled) it is folded into one sentence of the first.

`gofmt`, `go vet` and `go test ./pkg/htmx/...` clean.

PR #1026, opened 2026-08-16
fix(money): a numeric currency code answers with the same currency every time

### LINKED ISSUE #1025: GetCurrencyByNumericCode returns a random currency for numeric codes with more than one entry (840 → USD or USDSPACE)
## What happens

`money.GetCurrencyByNumericCode("840")` returns `USD` on some runs and `USDSPACE` on others, for the same input in the same binary.

`pkg/money/currency.go` holds two entries whose `NumericCode` is `"840"`:

```go
USD:      {..., Code: USD,      NumericCode: "840", Thousand: ",", ...},
USDSPACE: {..., Code: USDSPACE, NumericCode: "840", Thousand: " ", ...},
```

`USDSPACE` is a *formatting* variant, not a currency. The lookup walks the currency map, and Go map iteration order is randomised per process, so whichever of the two is visited first wins.

## Why it matters

`Currency.Code` is not just a formatting hint — consumers persist it. In EAI:

- `back/modules/insurance/.../portfolio_v2/admissions/settlement.go` maps 1C's ISO numeric code to the alphabetic one for the admissions drawer and its Excel export;
- `back/modules/reinsurance/infrastructure/commands/upsert-{contract,offer}/shared.go` (`NappCurrencyToCode`) writes the result into `reinsurance.contracts` / `reinsurance.offers` currency columns.

So a USD receipt can render as `USDSPACE`, and a NAPP-sourced reinsurance contract can be stored with `currency_code = 'USDSPACE'`, at random. (EAI prod carries no such rows yet — those columns are still NULL — but the write path is live.)

It also makes a downstream test red at random, which under EAI's release model blocks a release with no way to tell a flake from a break.

## Expected

`GetCurrencyByNumericCode` is deterministic and answers with the canonical ISO currency for a numeric code. Formatting variants like `USDSPACE` should not be reachable through a numeric-code lookup.

Suggested shape: build a `numericCode -> Code` index once from an explicit canonical list, rather than scanning the map; or exclude entries whose `Code` is not a 3-letter ISO code from the scan.

## Reproduce

```go
seen := map[string]int{}
for i := 0; i < 1000; i++ {
    seen[money.GetCurrencyByNumericCode("840").Code]++
}
fmt.Println(seen) // both USD and USDSPACE appear
```

diyor28:
Fixes #1025.

## The defect

`Currencies.CurrencyByNumericCode` ranged over the currency map and returned the first entry whose `NumericCode` matched. Go randomises map iteration, so a numeric code carried by more than one entry resolved differently between runs of the same binary.

Two codes are shared in the built-in table:

| Numeric | Entries | Nature |
| --- | --- | --- |
| `840` | `USD`, `USDSPACE` | `USDSPACE` is a thousands-separator variant, not a currency |
| `532` | `ANG`, `XCG` | a genuine succession in the standard |

`Currency.Code` is not only a formatting hint — consumers persist it. In EAI this reaches storage through `NappCurrencyToCode` (reinsurance contracts and offers) and through the 1C admissions drawer and its Excel export, so a dollar receipt could be written as `USDSPACE`. It also turned a downstream test red at random, which under EAI's release model blocks a release with no way to tell a flake from a break.

## The fix

The winner is chosen instead of stumbled upon: an entry whose `Code` is a three-letter ISO alphabetic code beats one that is not, and among those the lexicographically smallest `Code` wins. `840` therefore answers `USD`, and `532` answers `ANG` — the same on every call.

The rule applies to currencies a caller adds through `Add`, not only to the built-in table, so a downstream formatting variant cannot capture a numeric code either.

No signature changes, no behaviour change for the ~170 numeric codes that only one entry declares.

## Tests

Three added in `pkg/money/currency_test.go`: `840` resolves to `USD` across 200 lookups; `532` gives one stable answer across 200 lookups; and a `Currencies` built from a caller's own variant plus an ISO entry answers with the ISO one.

`GOWORK=off go test ./pkg/money/` is green.

---
<a href="https://app.blacksmith.sh/iota-uz/codesmith/iota-sdk/pr/1026"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789470450&installation_model_id=2042&pr_number=1026&repository=iota-uz%2Fiota-sdk&return_to=https%3A%2F%2Fgithub.com%2Fiota-uz%2Fiota-sdk%2Fpull%2F1026&signature=4aed7f3a4a7202d323e49a61034c5000b28a09bf112f60344c4527bb5cb6a35e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Currency lookups by numeric code now return consistent results when multiple currencies share the same code.
  * Standard three-letter currency codes are prioritized over formatting variants.
  * Runtime-added currencies follow the same deterministic selection rules.

* **Tests**
  * Added coverage for duplicate-code handling, stable results, and runtime-added currencies.

### DIFF
--- a/pkg/money/currency.go
+++ b/pkg/money/currency.go
@@ -18,15 +18,60 @@ type Currency struct {
 
 type Currencies map[string]*Currency
 
-// CurrencyByNumericCode returns the currency given the numeric code defined in ISO-4271.
+// CurrencyByNumericCode returns the currency given the numeric code defined in
+// ISO-4217.
+//
+// A numeric code is not unique in this collection. "840" is declared by both
+// USD and USDSPACE, where USDSPACE is a thousands-separator variant rather than
+// a currency of its own, and "532" by both ANG and XCG, which is a genuine
+// succession in the standard. Ranging over the map and returning the first hit
+// answers with whichever entry Go's randomised iteration happened to visit
+// first, so the same numeric code yields a different currency between runs of
+// the same binary - and callers persist Currency.Code, so that reaches storage.
+//
+// The winner is therefore chosen rather than stumbled upon: an entry whose Code
+// is a three-letter ISO alphabetic code beats one that is not, and among those
+// the lexicographically smallest Code wins. The answer never varies.
 func (c Currencies) CurrencyByNumericCode(code string) *Currency {
+	var best *Currency
 	for _, sc := range c {
-		if sc.NumericCode == code {
-			return sc
+		if sc.NumericCode != code {
+			continue
+		}
+		if best == nil || preferredCurrency(sc, best) {
+			best = sc
 		}
 	}
 
-	return nil
+	return best
+}
+
+// preferredCurrency reports whether candidate should displace incumbent as the
+// answer for a numeric code both of them declare.
+func preferredCurrency(candidate, incumbent *Currency) bool {
+	candidateISO, incumbentISO := isISOAlphabeticCode(candidate.Code), isISOAlphabeticCode(incumbent.Code)
+	if candidateISO != incumbentISO {
+		return candidateISO
+	}
+
+	return candidate.Code < incumbent.Code
+}
+
+// isISOAlphabeticCode reports whether code has the shape ISO-4217 gives an
+// alphabetic currency code: exactly three capital letters. Formatting variants
+// carried in this collection - USDSPACE and anything a caller adds like it -
+// do not, which is what keeps them from answering a numeric-code lookup.
+func isISOAlphabeticCode(code string) bool {
+	if len(code) != 3 {
+		return false
+	}
+	for i := 0; i < len(code); i++ {
+		if code[i] < 'A' || code[i] > 'Z' {
+			return false
+		}
+	}
+
+	return true
 }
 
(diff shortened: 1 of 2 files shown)


diyor28:
Тест на 532 теперь утверждает ANG явно (2bb10b7): оба кода ISO-формы, тай-брейк — меньший код, поэтому ожидаемый ответ детерминирован и не зависит от первого прогона.

Issue #957, opened 2026-07-30
Publish SHA-stamped frontend packages and finalize Go-npm pins atomically


diyor28:
## Problem

iota-sdk Go commits and frontend package versions can drift across consumers.
Human-maintained semver lockstep is not sufficient for a platform whose Go
descriptors/contracts and JavaScript runtime must change atomically.

## Scope

- Produce immutable SHA-stamped artifacts for every publishable frontend
  package from one iota-sdk commit.
- Support local/PR preview through Go module overlay plus exact npm tarballs.
- Extend consumer finalization to update Go and npm pins from the same SDK
  commit in one change.
- Record package/protocol provenance so `doctor` can diagnose mismatches.
- Prove five consecutive automatic publish/finalize cycles before enabling an
  exact-match enforcement gate.
- Preserve independently importable package boundaries within one release
  train.

## Out of scope

- One monolithic npm package/bundle.
- EAI feature migration.
- Unrelated CI optimization.
- Automatic breaking-change adoption without preview.

## Acceptance

- [ ] Each SDK commit yields immutable, traceable frontend artifacts.
- [ ] A consumer preview resolves matching Go and npm sources automatically.
- [ ] Finalization updates both dependency planes atomically.
- [ ] Partial publish/finalize failure cannot leave a consumer on mixed
  commits.
- [ ] Exact-match enforcement remains disabled until five consecutive
  unattended cycles succeed.

diyor28:
Architecture update (implemented in #984): **bounded internal modules, one public npm distribution with stable subpath APIs**.

`web/client-host` and `web/lens` remain separate private workspace modules. Production publishes only `@iota-uz/sdk`, with `./identity`, `./client-host`, `./lens`, and `./lens/styles.css`. PR previews are one ephemeral Actions artifact and never registry dependencies. Go tag, npm version, source commit, and protocol identity are validated as one release unit.

Per the latest product decision, 0.5 is a clean break: legacy root/BiChat/applet/Tailwind/assets and the accidental `@iota-uz/client-host` / `@iota-uz/lens-web` packages are not compatibility surfaces.

Issue #953, opened 2026-07-30
Add shell widget slots and shared React portal services


diyor28:
## Problem

Persistent shell widgets and React overlays currently share no ownership
contract. Lens and applets can append body-level portal roots, duplicate theme
state, compete on z-index, lock body scrolling independently and manage inert
and focus themselves.

## Scope

- Define named shell widget slots with explicit mount/unmount and document
  lifetime.
- Define separate portal services for modal, drawer, toast and command-palette
  surfaces.
- Centralize portal roots, z-layer ordering, theme propagation, focus restore,
  focus trap, inert/background behavior and scroll locking.
- Define collision/nesting rules and accessibility requirements.
- Provide renderer-specific adapters for Templ shell services and React
  consumers without allowing HTMX mutation inside React roots.
- Migrate one Lens overlay as the reference consumer.

## Out of scope

- Making the server shell persistent across document navigation.
- Arbitrary React islands inside server page content.
- Replacing feature-owned application state.
- Ali migration.

## Acceptance

- [ ] Widget slots and portal services are separate public contracts.
- [ ] Features do not create their own body-level modal/drawer roots.
- [ ] Nested overlay order, focus, inert and scroll lock are deterministic.
- [ ] Theme and accessibility behavior are identical in light/dark modes.
- [ ] The reference Lens overlay uses the host service and removes its
  duplicated body/theme lifecycle.

Issue #952, opened 2026-07-30
Unify Granite design tokens and semantic UI states across Templ and React


diyor28:
## Problem

Templ/sfui and React must look like one Granite product without pretending that
their component APIs or lifecycles are portable between frameworks.

## Scope

- Define the canonical CSS custom-property contract for color, typography,
  spacing, radius, shadow, motion and z-layers.
- Consume the precompiled Tailwind/theme distribution from #834.
- Publish the same tokens to server components and the bounded React UI
  package.
- Define shared semantic states: loading, empty, error, disabled, selected and
  destructive.
- Establish a small parity inventory for primitives and application patterns;
  explicitly mark renderer-specific components.
- Prohibit feature-local global resets, fonts, primary palettes and independent
  theme models.
- Add representative light/dark/accessibility fixtures for both renderers.

## Out of scope

- Porting every Templ component to React.
- One cross-framework component API.
- Redesigning existing product pages.
- React ResourceTable/Form.

## Acceptance

- [ ] One token source drives sfui/Templ and React.
- [ ] Both renderers use one typography and light/dark contract.
- [ ] The parity inventory is enumerated rather than implied.
- [ ] Renderer-specific escape hatches are documented.
- [ ] Reference server and React pages pass the same semantic-state and
  accessibility checks.

Issue #949, opened 2026-07-30
Add renderer-aware route descriptors and a standard React route host


diyor28:
## Problem

`application.RouteSpec` can describe method/path/auth metadata but cannot
declare whether a route is server- or client-rendered. Applet route descriptors
carry no auth metadata, and their iota-sdk adapter therefore cannot be the
security authority for a standard internal React route.

Client features also register mounts, catch-all handlers, asset tags, and
frontend entries independently.

## Scope

- Add an explicit `Server | Client` renderer disposition to the canonical Go
  route descriptor.
- Consume the non-nil access Requirement from RBAC v2 (#868), with explicit
  public, authenticated, and permissioned constructors.
- Fail composition when renderer or access is unspecified.
- Define native client-route metadata: stable route ID, path/prefix,
  navigation identity, and frontend feature ID.
- Register client routes through one standard server-shell handler using the
  extracted iota-sdk runtime (#948).
- Generate an inspectable route-to-lazy-module manifest; TypeScript must not
  maintain a second path or permission registry.
- Preserve route splitting: a server page may link to a client-owned sub-route
  under the same navigation identity.
- Validate duplicate routes and full route/access catalogs at composition
  time.

## Migration boundary

Legacy applet descriptors remain supported through compatibility wrappers but
are not treated as authoritative access declarations. Existing server
controllers continue to work and can adopt explicit renderer metadata
incrementally.

## Out of scope

- Bootstrap/session/i18n/portal service contents.
- RPC transport hardening.
- EAI workspace/build implementation.
- Product route migrations.
- React islands inside HTMX-owned subtrees.

## Acceptance

- [ ] Every newly declared route has an explicit renderer and non-nil access
  Requirement; omitted values fail startup.
- [ ] A client route is registered from one Go descriptor and resolves one lazy
  frontend feature without a handwritten mount or TypeScript route copy.
- [ ] Navigation visibility and HTTP access consume the same Requirement.
- [ ] Client routes render inside the existing authenticated server shell.
- [ ] Route splitting works without mixed DOM ownership.
- [ ] Integration tests cover server/client routes, public/authenticated/
  permissioned access, duplicate detection, and legacy applet compatibility.

PR #1077, opened 2026-09-14
feat(web): add Solid UI parity and Go-owned screens

diyor28:
## Summary

The Solid runtime now has both visual parity with the public templ component surface and a Go-owned feature API. A feature package declares a colocated TSX screen with `solid.Import`, declares typed queries and actions in the same Go value, and renders it through the normal client-host route without a hand-written frontend registry or mount function.

`iota-solid` discovers those declarations statically and commits two reviewable outputs: a binding beside the TSX source and one lazy application catalog. Its check and watch modes reconcile add, move, and delete operations; unsupported wire types, duplicate features or RPCs, missing sources, and incompatible component exports fail before runtime.

The client host owns route loading, safe bundle errors, Retry, theme propagation, mount, and idempotent cleanup. A feature or controller chooses one renderer for its owned UI: Solid or templ + HTMX. Sharing the server shell is supported; mixing renderer ownership inside a feature remains an explicit transition exception.

## Solid UI parity

`@iota-uz/sdk/solid-ui` provides a Solid component or typed adapter for all 185 public templ component exports. It includes forms, data display, selection, navigation, overlays, table and scaffold workflows, import and multilang flows, sidebar and shell primitives, charts, slots, async utilities, browser and SSR entry points, canonical CSS, and package fonts.

The Playwright visual lane covers light and dark themes, desktop and mobile layouts, interactive states, DOM and accessibility contracts, and exact comparisons against Go-templ fixtures.

## Developer experience

```go
var productCreate = solid.Import[ProductCreateProps](
    "./web/product-create/ProductCreate.tsx",
    solid.WithRPC(
        solid.Query[ReferencesRequest, References]("references", solid.Cacheable(2)),
        solid.Action[CreateRequest, CreateResponse]("create"),
    ),
)

func (c *Controller) Key() string {
    return application.ClientRoute("insurance.product-create")
}

// In the normal client-host payload:
Screen: productCreate.Render(props)
```

Generated bindings give the TSX module typed props and RPC calls. Production startup performs no generation and requires no Node runtime.

## Review follow-up

The review fixes cover async Combobox form submission, clearable and strictly parsed money values, dynamic uncontrolled Select values, document-scoped floating placement with resize observation, contextual host error logging, deterministic import paths, renderer consistency per controller, and portal/theme lifecycle edge cases. `web/solid-ui/PARITY.md` was removed; executable tests and generated contracts are the maintained source of truth.

## Validation

- `go test ./internal/solidgen ./pkg/clienthost/... ./pkg/application/... ./cmd/iota-solid/...`
- `go vet ./...`
- `pnpm --dir web/client-host typecheck`
- `pnpm --dir web/client-host test` — 44/44
- `pnpm --dir web/solid-ui test` — 137/137
- `pnpm --dir web/client-host test:browser` — 3/3
- `pnpm --dir web/sdk build && pnpm --dir web/sdk check`
- Solid UI visual regression — 224/224 local exact comparisons; baselines are intentionally not committed as PR artifacts
- `git diff --check`

The repository-wide Go test command also reached unrelated integration packages; the local environment lacks the expected `iota_erp` PostgreSQL database and Python kernel runtime. All changed Go packages pass in the focused command above.

Refs #947
Refs #949
Refs #952
Refs #953
Refs #957


diyor28:
@coderabbit review

diyor28 (commented):
## Deep review — traced every code path

Worktree: `33b4904` vs base `9236406` (PR base, not `origin/main` tip — `origin/main` has moved since). Read `generator.go` (758), `watch.go`, `solid.go` (191), `wire.go` (148), `host.go` delta, `solid.ts`/`solid-portals.ts`/`portal-host.ts`/`portals.tsx` deltas, `solid.test.ts`/`portal-host.test.ts`, sampled `solid-ui` (`Select`, `Input`/`MoneyInput`, `Combobox`, `Toast`, `floating.ts`, `index.ts`, `PARITY.md`, packaging scripts, `iota.css`), CI/Justfile/docs/ADR. Ran `pnpm --dir web/client-host test` — 43/43 pass (incl. 11 solid + 2 portal-host). Go tests could not run here (module cache volume read-only), so Go findings are static-trace only.

Overall: high quality. Generator fail-fast surface, two-pass safe-int validation, lease-stacked theme, ref-counted portal locks, and bounded bundle-error UI are all well reasoned. No security hole or protocol break found. Findings below are ordered by severity; nothing here demands a merge block since `solid-ui` is new and not yet serving prod routes, but 1–3 deserve a fix or a tracked follow-up before consumers rely on them.

---

### Major

**1. `Combobox` async selections never reach native form submission** — `web/solid-ui/src/selection/Combobox.tsx:177-179`
The hidden `<select>` that carries `name`/`form` renders `<For each={local.options}>` only. When options come from `loadOptions`/`endpoint`, `sourceOptions()` is `remoteOptions()`, and `selectedOptions()` correctly falls back to `{value, label: value}` — but the hidden select has no matching `<option>`, so the browser submits empty. Either render `sourceOptions()` + selected-fallback into the hidden select, or document that async mode requires `onValueChange` + manual submission (and then the hidden select is misleading and should be omitted in async mode).

**2. `MoneyInput` cannot be cleared; multi-decimal input silently truncates** — `web/solid-ui/src/forms/Input.tsx:139-143,170`
`parseMoney("")` → `parseFloat("")` → `NaN` → returns `0`. Clearing the field snaps value to 0 and re-formats to e.g. `0.00`, so users can never empty it. Also `"1.2.3"` normalizes to `"1.2.3"` and `parseFloat` yields `1.2` with no error. Suggest: return `NaN`/empty sentinel and propagate empty string to display while emitting `0` or `undefined` on the hidden input explicitly; validate single-decimal and surface a validation error otherwise.

**3. `createFloatingPlacement` is bound to global `document`/`window`** — `web/solid-ui/src/internal/floating.ts:27-28,40-46`
`viewportWidth/Height`, `resize`, and capture-phase `scroll` all use globals, while the rest of this PR (`portal-host.ts`, `solid.ts`) deliberately moved to `owner.ownerDocument` for iframe/test correctness. Breaks in iframes, multi-document tests, and SSR (`window` undefined; there is an `ssr-smoke.mjs` but it is 7 lines — worth confirming it imports this module). Also no `ResizeObserver`, so sidebar collapse / layout shift leaves placement stale until scroll/resize. Suggest threading `anchor().ownerDocument.defaultView` with a null-guard.

### Medium

**4. `solid.Import` wrapper indirection silently forks FeatureID** — `pkg/clienthost/solid/solid.go:85-112` + `internal/solidgen/generator.go:234-303`
Runtime ID comes from `runtime.Caller(1)` package (`//go:noinline` only protects one frame); generator ID comes from the textual `pkg.PkgPath`. A helper like `func MyImport[P](s string) solid.Component[P] { return solid.Import[P](s) }` compiles and passes `validateSource`, but runtime attributes the feature to the helper's package while the generator attributes it to the caller's package. `host.go:196-199` would then set `Spec.FeatureID` to the runtime value while the catalog key is the generator value → runtime `absent from the generated catalog`. No code prevents this; suggest a doc comment on `Import` ("do not wrap") plus a generator-side note, or a `-check` cross-validation in CI that imports the package and compares `FeatureID()`.

**5. Handler error observability: every failure is a bare 500** — `pkg/clienthost/host.go:239-306`
Build error, Screen nil/mismatch, `validateSafeIntegers` failure, `json.Marshal` failure, session failure, shell render failure all return `http.Error(500)` with no log, no route/feature ID, no distinction. The safe-int path in particular (`wire.go` produces precise `clienthost props $.items[3].value: ...` messages) discards the message that would make debugging possible. Pre-existing pattern, but this PR adds three new 500 branches; suggest at minimum a structured log with `controller.id`, `route.Spec.RouteID`, and the validation error (never echo props values).

**6. `wire.go` conservative rejections worth documenting** — `pkg/clienthost/wire.go:96-104,33-57`
(a) `,string` fields and `encoding.TextMarshaler` values are validated on the Go numeric value even though the wire form is a quoted string — false positive for large ints intentionally string-encoded. (b) Any JSON number with `|n| > 2^53-1` is rejected, including legitimate floats like `1e21`. Both are defensible (fail-closed), but the doc comment only mentions integers; extend it to state floats/`,string` are intentionally covered so a future author doesn't "fix" it into an unsafe hole.

**7. `Select` uncontrolled `defaultValue` is lost on option refresh** — `web/solid-ui/src/forms/Select.tsx:47-55`
The `ca4e815` fix re-applies `local.value` when `resolvedChildren()` changes, but the `onMount`-only `defaultValue` branch never re-runs. After an async option refresh, an uncontrolled select snaps to the first option. Either re-apply `defaultValue` when `value === undefined` inside the effect, or document controlled-only for dynamic options.

**8. ADR one-renderer rule is prose-only** — `adr/platform/solid-features-use-go-import-and-generated-catalog.md:28-29`
"Feature uses Solid entirely or templ+HTMX entirely; sharing the server shell is supported; mixing is an explicit transition exception" has no enforcement point I could find (`grep` for renderer mixing checks; `host.go` validates per-route, not per-feature). Fine as a guideline, but either add a lint (e.g. `iota-solid` warns when a package with `solid.Import` also declares templ handlers for the same route prefix) or reword to "convention".

### Minor / nits

- **Dead code**: `ExportName` (`internal/solidgen/generator.go:752`) is exported but unreferenced anywhere (`git grep` confirms). Delete or use.
- **Dead statement**: `web/client-host/src/solid.ts:186` (`throw cause` after `throwCleanupFailures`, which always throws since `failures` is non-empty). Harmless but confusing; remove.
- **Generator/option strictness mismatch**: `resolveMethods` requires `Invalidates` targets to be string *literals* (`generator.go:361-371`), while runtime `Invalidates(...string)` accepts consts. Intentional for static analysis, but a const-driven codebase will hit a confusing generator-only error. Mention in the `Invalidates` doc comment.
- **Non-clean source accepted**: `"./a//b.tsx"` passes both `scanFile` (`generator.go:220-224`) and `validateSource` (`solid.go:182-191`) because neither enforces `source == "./"+path.Clean(trimmed)`. `FeatureID` cleans, and `filepath.Join` cleans the output path, so both sides agree — but the committed binding filename derives from the cleaned path while the source text is unclean. Reject non-clean inputs for determinism.
- **`json:"-,"` edge**: `jsonField` (`generator.go:640-661`) treats any tag whose first segment is `-` as skip, but Go's `json:"-,"` means field named `"-"`. Pre-existing trivia; only matters for a field literally named `-`.
- **Watch gaps**: `watch.go:57,82-84` — `Rename` (covers `mv`/checkout-created dirs) never triggers `addWatchDirectories`, only `Create` does; `report` runs synchronously and can stall the event loop if the consumer blocks. Also note the good: generated `*.generated.ts` (not `.tsx`) correctly avoids self-trigger loops.
- **Portal root side effect during render**: `solid-portals.ts:39` calls `host.registry.root()` (which creates+appends the surface root) during component render, not in `onMount`. If a later line throws, an empty `[data-iota-portal]` div leaks. Move the `root()` call into `onMount`.
- **`WidgetSlot` silent no-op**: `solid-portals.ts:105-108` returns `undefined` when outside a host or when the slot is absent — misconfiguration renders nothing with no signal. A `console.warn` (or throw for missing host, matching `Portal`) would save debugging time.
- **Toast event names**: `overlays/Toast.tsx:90-94` defaults to `'notify'` / `'remove-notification'` on `defaultView`. `'notify'` is collision-prone (other libs, product code). Namespace to `'iota:toast-notify'`-style, or at least document the collision domain.
- **React/Solid theme-restore asymmetry**: Solid `setPortalTheme` (`solid.ts:53-71`) preserves/restores base via lease stack (covered by the out-of-order dispose test — nice); React `ClientHostProvider` (`portals.tsx:21-24`) blindly `delete`s `dataset.theme` on cleanup. Harmless today (Go bootstrap leaves portal owner themeless), but a product shell that pre-themes the portal node will lose it under React and keep it under Solid.
- **`documentTheme` reads `classList`, Go writes `data-theme`**: `solid.ts:224-230` vs `host.go:348`. Works via the `fallback` today, but a shell that only sets `data-theme="dark"` on `<html>` (like `StandaloneShell`) with no `dark` class will be ignored by the client override. Read `data-theme` as an additional signal.
- **Error UI keeps `solid-loading` ownership** (`solid.ts:254,285-307`): intentional since Retry does a full reload (correct — rejected dynamic imports are cached per document), but a custom `reload` that doesn't navigate leaves the root permanently owned. Document that custom `reload` must navigate or call `dispose`.
- **`iota.css` fixes are correct** (`styles/tailwind/iota.css`): `--crl-`→`--clr-`, `--transaprent`→`--transparent`, `.dialog-tbd`→`.dialog-ttb`, and `color: oklch(var(--clr-form-control-label))` all match the file's channel-variable convention (`--gray-400: 70.7% 0.022 261.325`). Note this *is* a visual change (previous vars were guaranteed-invalid → fell back), so VR diffs on buttons/labels/dialogs are expected, not regressions.
- **Lens bundle cap 439k→440k** (`web/lens/src/charts/bundle.test.ts`): justified in-comment (shared portal host, ~1k). Tripwire preserved.
- **VR baselines**: only `vr/baselines/linux/README.md` is committed; the 224/224 claim isn't verifiable from the tree — CI mints candidates as artifacts for promotion. Expected workflow, but call out in the PR description which CI run holds the candidate artifact for this SHA so reviewers can inspect.
- **Packaging**: `web/sdk` subpath assembly (`assemble-subpaths.mjs`, `check-package-exports.mjs`) correctly asserts package-relative font URLs and required font files — good. `solid-ui` stays `private:true` and ships via the SDK dist (no separate publish footgun). Confirm `pnpm --dir web/solid-ui install` in CI resolves against the workspace root lockfile as intended (it does today, but the step name says "Install Solid UI dependencies" while it effectively installs the workspace).

### Praise (to lock in)

- Two-pass integer safety (`walkSafeIntegers` + `walkJSONNumbers` over the marshaled form) with cycle protection and `json.Marshaler` bypass covered by `TestController_RejectsUnsafeNumbersFromCustomJSONMarshalers` — the right shape for a JS-number boundary.
- Generator ownership model: header-marker stale detection that preserves user-owned `*.generated.ts` without the marker (`TestRunCheckReportsDrift...`), deterministic sorted outputs, `-check` for CI.
- `PortalRegistry` extraction: `ownerDocument`-scoped roots, ref-counted document/background locks, nested-restore preservation (`portal-host.ts:77-85`), and the shared-host test in `portal-host.test.ts` directly cover the React+Solid coexistence this PR introduces.
- Catalog loader: rejected-import vs invalid-module error split is correct (`then(onF, onR).catch`), Retry reloads (required by import-cache semantics), late-resolve-after-dispose is ignored via generation counter — all covered in `solid.test.ts`.
- `PARITY.md` honesty: `COMPOSED`/`ADAPTER` explicitly count as mapped *without hiding* reduced behavior, and the subscription-package scope exclusion is stated. Spot-checked `index.ts` surface — comprehensive; consider a CI assertion that templ-export count vs mapping stays in sync so the table can't rot.

diyor28:
Addressed the deep-review findings in `dc8d74904` and the lint follow-up in `17476ea6b`.

Implemented:

- async Combobox values now participate in native form submission;
- MoneyInput supports empty values, rejects malformed decimals, and blocks invalid native submission;
- floating overlays use their owning document and observe layout-size changes;
- uncontrolled Select preserves the current value across option refreshes;
- client-route failures emit contextual structured logs;
- conservative numeric-wire behavior and static-literal requirements are documented;
- clienthost controllers reject mixed React/client renderer ownership;
- dead generator/runtime code was removed, import paths are canonical, `json:"-,"` is handled correctly, rename watches reconcile directories;
- Solid portals create roots only after mount, missing widget slots are visible to developers, React restores the shell theme, and `data-theme` participates in Solid theme discovery;
- custom Retry reload ownership is documented;
- `web/solid-ui/PARITY.md` was removed as requested.

Two suggestions were intentionally not applied:

- toast events remain `notify` / `remove-notification` because those names are the existing templ/HTMX interoperability contract;
- watcher reports remain synchronous because the only production consumer is the bounded CLI reporter; making callbacks concurrent would introduce ordering and shutdown ambiguity for no current benefit.

Validation on head `17476ea6b2506f61b9f4913b11d8e01ff3b6e8a0`: all 9 SDK checks pass. Granite PR #4687 also passed `SDK preview build` against that exact SHA.

diyor28 (commented):
## Second pass — all 16 findings re-traced at `17476ea`

Scope: `33b4904..17476ea` (`dc8d749` + `17476ea`). Verified with `go vet` (clean), `go test ./internal/solidgen/ ./pkg/clienthost/...` (all ok), `pnpm --dir web/client-host test` (44/44, +1), `pnpm --dir web/solid-ui test` (137/137 across 18 files, incl. new floating/forms/selection tests). Go run via `CGO_ENABLED=0` + temp module cache (sandbox cache volume is read-only; unrelated to PR).

Verdict: everything from the first pass is correctly resolved. Three majors are genuinely fixed with regression tests, not papered over. Details + a few new minor observations below.

---

### Resolution check

**1. Combobox async submit — fixed.** `formOptions()` (`Combobox.tsx:79-86`) unions source options with selected-value fallbacks and feeds the hidden `<select>`. The `selection.test.tsx` FormData assertion (`get('customer') === '42'`) covers exactly the reported hole. `selectedOptions()` fallback at line 93 is now consistent with what gets submitted.

**2. MoneyInput clear/truncate — fixed.** Tri-state `parseMoney` (`undefined` empty / `null` malformed / cents) + `draft` overlay so typing is never reformatted mid-keystroke, `setCustomValidity` on invalid, hidden input emits `''` when empty, `describedBy` now keys off actual `validationError()` presence (the old min/max-presence reference pointed at a nonexistent node — quietly fixed too). Test covers clear-then-`1.2.3`. Two deliberate notes: `onValueChange` widened to `number | undefined` (breaking for any early adopter — fine pre-release, but call it out in the PR description), and a controlled parent updating `value` while a draft exists loses to the draft until blur — the right choice (don't clobber typing).

**3. Floating globals — fixed.** Viewport, resize/scroll listeners, and the `ResizeObserver` constructor all resolve from `anchor.ownerDocument.defaultView`, with null-guards for SSR/test documents. Refs are assigned before `onMount` in Solid, so the mount-time listener capture at `floating.ts:62-66` sees the real anchor. The iframe test pins `align: 'end'` against a 300px owner viewport while the global document is 1000px — directly falsifies the old behavior. `Promise.resolve().then(update)` vs `queueMicrotask` is equivalent here.

**4. Import wrapper fork — documented, accepted.** `Import` ("call it directly… wrappers cannot be discovered") and `Invalidates` (literals-only) doc comments state both constraints at the declaration site, which is where a wrapper author will read them. Runtime fork remains theoretically possible but is now an explicit contract violation, and `-check` fails loudly on the generator side.

**5. Bare 500s — fixed.** `fail()` logs `controller.id`/`route.id`/`route.path`/`stage` + cause; messages never echo props. Test asserts all three fields plus the cause string. I confirmed no loop-var capture (`Register` copies `route` per index, `handler` takes it by value) and that `WithLogger` nil-guards with a `StandardLogger` default. One follow-up: no prod call site passes a logger yet (`WithLogger` only used in `host_test.go:97`) — wire the app logger when controllers get constructed in bootstrap, or these land in the default logger stream.

**6. wire.go conservatism — documented.** Comment now names `,string`, text-marshaled numerics, and out-of-range floats explicitly. Correct scope for a fail-closed boundary.

**7. Select defaultValue — fixed, and better than asked.** `uncontrolledValue` signal + effect re-assign on options/children change + `onChange` forwarding. Regression test refreshes options mid-selection. Behavior nuance worth knowing: uncontrolled with no default and no placeholder now renders blank (`selectedIndex = -1`) instead of silently pre-selecting the first option — that's a fix, not a regression (no more unintended submitted values), and the placeholder pattern (`value=""`) keeps working through `assignValue('')`.

**8. ADR renderer rule — enforced, not just reworded.** `NewController` rejects mixed client/react sets with a clear message + `TestController_RejectsMixedRenderers`, ADR text updated to match (per-controller enforcement; cross-controller boundary stays a review contract). Verified `NewController` is only constructed in tests in-tree, so nothing existing trips the new rejection; ordering is right (renderer allowlist check precedes seed/compare).

**Nits — all resolved:** `ExportName` deleted with its `unicode` import; dead `throw cause` refactored into `cleanupFailure` with identical semantics (`[cause]` → rethrow original, preserving stack); non-clean sources rejected on both Go sides with tests (`generator_test.go` non-clean case, `solid_test.go` `./a//` + `./a/..` cases); `jsonField` now matches `encoding/json` exactly (`tag == "-"` skips; note this also means `json:"-,omitempty"` is *not* skipped — the old code was stricter than stdlib, the new code agrees with it; test pins both `-` and `-,`); watch handles `Rename` (generated `*.generated.ts` still can't self-trigger); `Portal` builds `root()` in `onMount`, reads `ownerDocument` via the new registry getter — render-phase side effect gone; `WidgetSlot` throws without host (was silent no-op — correct behavior change) and warns with the slot name; `documentTheme` reads `data-theme` first with a test; `reload` documents the navigate-or-dispose contract. Toast `notify`/`remove-notification` defaults unchanged, but I retract the nit: both are overridable props (`notifyEvent`/`removeEvent`) and `overlays.test.tsx` pins the wiring — configurable default is fine.

### New observations (all minor)

- **PARITY.md deletion is justified — with one coverage caveat.** `playwright.config.ts:42-54` starts both the Solid gallery and `go run ./vr/templ-fixture`, `testDir: './vr'` picks up `templ-parity.spec.ts`, and CI runs it via `just solid-ui vr` — the static 185-row table is replaced by a live geometric contract, which is strictly better. Caveat: the direct templ-vs-solid cases cover a handful of components (button/input/checkbox/select/radio); the rest of the "all mapped" claim rests on gallery screenshots (threshold 0, `maxDiffPixels: 0` — admirably strict). Consider growing `templ-parity.spec.ts` toward high-risk components (Combobox, MoneyInput, table workflow) as they stabilize.
- **Watch `Rename` cost.** Every rename (editor atomic saves, `git checkout`) triggers a full root re-walk of `watcher.Add` calls, undebounced — only `Run` is debounced. Dev-tool-only and correct, but gating the re-walk on `relevantGraphFile` misses or debouncing it would avoid syscall storms on large checkouts. Trivial.
- **Combobox `endpoint` transport.** `fetch(url, { signal })` sends same-origin cookies by default (fine for session auth) but offers no header injection — CSRF-token headers require `loadOptions`. One line in the `endpoint` doc comment saying so would pre-empt a confused integration.
- **`Select` dead branch.** `assignValue`'s array path is unreachable (`value`/`defaultValue` are typed `string`). Harmless; delete if touched again.
- **jsdom gap (acknowledged, not a defect).** `floating.test.tsx` can't exercise the `ResizeObserver` path (absent in jsdom → early return). The observer callback reads live accessors and only re-observes on element identity change, so no feedback loop is possible by construction — fine without a browser test.

No blockers. The three majors closed with tests, the enforcement nit became real enforcement, and the suites I can run are green.

